PR #1, opened 2016-09-01
Update install-configure-admiral.md

asual:
The Admiral UI no longer requires /uic and is now available in the root path.


tgeorgiev:
I think we have to discard or postpone this, Admiral UI will have to stay at /uic for as there are certain services that don't work well with the root path. We will follow up on this soon

tgeorgiev:
Update, the problem has been resolved, now the latest release https://github.com/vmware/admiral/releases/tag/v0.5.0 makes the Admiral UI available at the root path, please continue with merging this pull request.

mlh78750:
Oops. I made this change already before seeing this PR. Closing it as the change has been made.

PR #3, opened 2016-09-27
Updated pointers to the 0.6.0 release + added instructions re how to …

mreferre:
I have augmented the documentation by adding info on how to pull images from Harbor into a VCH. In addition I have also fixed the VIC version to the latest (0.6.0) including download links.


mlh78750:
LGTM

docs/setup/alpha/using-harbor.md
```diff
@@ -54,11 +56,34 @@ We are now pushing nginx:1.9 to Harbor in the _vmworld_ project (we can do that
...
 1.9: digest: sha256:311e9840c68d889e74eefa18227d0a6f995bc7a74f5453fdcd49fe3c334feb24 size: 1978
 ```
 
 If you check the Harbor UI, in the vmworld project, you should now see a new item (i.e. nginx:1.9)
+
+#### Interacting with Harbor with a Virtual Container Host (VCH)
+
+Similarly, we can configure a VCH to interact with a local Harbor registry. Should you have proper certificates in place, the interaction is going to be fairly similar to what we have seen above.
+
+In case you need to consume a Harbor instance that has self-signed certificates, then you need to set the `--insecure-registry` at the VCH level.
```
sflxn: VIC doesn't support private secure registries yet.

docs/setup/alpha/using-harbor.md
```diff
@@ -54,11 +56,34 @@ We are now pushing nginx:1.9 to Harbor in the _vmworld_ project (we can do that
...
+
+To do so you have to deploy the VCH specifying the FQDN/IP of the insecure registry you will connect to.
+
+For example, you can add `--insecure-registry 10.140.50.77` to the `vic-machine` command you use to deploy the Virtual Container Host. This will tell the VCH that it's ok to pull from the Harbor registry hosted at 10.140.50.77.
```
sflxn: The option is actually --docker-insecure-registry.  Also, to add more than one, they just repeat the option for each IP.

docs/setup/alpha/using-harbor.md
mlh78750: Thanks @sflxn. I missed this.

mreferre: @sflxn thanks for chiming in. Can you clarify what we mean by "private secure registries"? I am asking because as part of a parallel "project" we are testing how third party registry (Vs Harbor) and how third party container managers (Vs Admiral) work with VIC Engine and one of the tests I have completed is by pulling from quay.io and bintray.com. 

They are obviously not "private" registry but if by "private" we mean "everything except the docker hub" than I found VIC Engine to be able to work out of the box with those registries (which have a proper certificate). /cc @mlh78750

docs/setup/alpha/using-harbor.md
mreferre: I will fix the typo. Thanks.

docs/setup/alpha/using-harbor.md
sflxn: Private registry as in a registry private to the user organization.  Secure just means communication with it is secure (https).  We're adding it so we should be able to test it at some point.

mreferre: Thanks @sflxn. I thought when I was connecting to quay.io I was indeed using a secure connection (I didn't add any --docker-insecure-registry flag to the VCH deployment string and yet I was able to pull an image from it. So what would the difference be between quay.io (which I assume uses proper certificates and an on-prem registry setup that uses proper certificates? That's the point i am missing.

Thanks.

mlh78750: I agree with @mreferre. If public secure works, what would prevent local secure? How would the client/VCH even know?

sflxn: There is the scenario where a secure private registry uses self generated certificates.  We don't have a way to upload those certificates to the VCH yet.  Also, we're not yet verifying CA issues certificates yet.  These should be resolved soon, but I just wanted to bring attention to the fact that they aren't working properly yet.

docs/setup/beta/install-configure-vch.md
```diff
@@ -57,14 +57,14 @@ Just like a Docker host, a VCH has a local image cache and can create and store
 
 - If deploying to a vSphere cluster, you need to have DRS enabled
 - The network PortGroups the VCH connects to need to be on a distributed virtual switch
```
hmahmood: A distributed switch is only required if deploying to VC. Standard switch suffices for a standalone ESX deployment

docs/setup/beta/known-issues-limitations.md
```diff
@@ -10,35 +10,35 @@ This means that if you pull an image into a VCH image cache using a particular p
 
 Deploying containers to a VCH deployed in nested ESX with some forms of shared storage can lead to timeouts. This is more to do with storage latencies which is exaccerbated by the nesting. iSCSI NAS seems to work fine, but NFS performance can be limiting. This is currently being investigated. https://github.com/vmware/vic/issues/1822.
 
-###Mapping port 8080 is not supported with VIC Engine 
+###Mapping port 8080 is not supported with VIC Engine
```
mreferre: thanks @hmahmood I haven't yet looked at the know issues / limitations file. I wanted to go through them one by one later tomorrow. One less :) Thanks.

docs/setup/alpha/using-harbor.md
hainingzhang: For the OVA version of Harbor, there is currently no way for a user to upload a certificate, neither self-signed nor public CA signed. That said, VCH can only connect to Harbor via HTTP, at least for the beta. In GA, we plan to allow users to upload a certificate.

mreferre: Good catch @hainingzhang. I am re-wording `in case you need to consume a Harbor instance that has self-signed certificates, then you need to set the --docker-insecure-registry at the VCH level` into `in case you need to consume a Harbor instance via plain http, then you need to set the --docker-insecure-registry at the VCH level`. Would that work? 

P.S. Out of curiosity (for the VIC Engine team): when I will be able to upload a non signed certificate into Harbor and access it via https.... how do I make the VCH accept the fact that the certificate isn't a real one (but rather a self-signed certificate)? Should I still use the `--docker-insecure-registry` flag? Just curios.

mlh78750:
I merged all these changes with #5.

Issue #49, opened 2017-03-21
Documentation: clean up Harbor topics


stuclem:
_From @stuclem on March 6, 2017 8:53_

The topic https://vmware.github.io/vic-product/assets/files/html/0.8/vic_installation/deploy_vch_harbor.html was thrown together in a hurry at the last minute and definitely needs an overhaul. At the least it needs to be more procedural.

The updated topic should be back-ported into the 0.8 docs.

_Copied from original issue: vmware/vic#4142_

stuclem:
The following topics need overhaul:

- [Deploy a VCH for Use with vSphere Integrated Containers Registry](vic_vsphere_admin/deploy_vch_registry.md)
- [Using vSphere Integrated Containers Engine with vSphere Integrated Containers Registry](vic_app_dev/using_vic_with_registry.md)

Issue #22, opened 2017-02-13
Document the unified OVA installer


stuclem:
Per https://github.com/vmware/vic/issues/3917.

stuclem:
See also https://github.com/vmware/vic/issues/3912.

stuclem:
Skeleton OVA is implemented in PR https://github.com/vmware/vic/pull/4041.

stuclem:
See https://github.com/vmware/vic/issues/4256 for Admiral integration and https://github.com/vmware/vic/issues/4257 for Harbor.

stuclem:
VIC Engine UI install will also be part of the unified installer, per https://github.com/vmware/vic/issues/4254.

stuclem:
Based on some conversations with @frapposelli and @martin-borisov, and having deployed a feature-incomplete version of the OVA, this is a summary of how it works:

- Download OVA from vmware.com or other hosting site
- Deploy OVA, providing Harbor/Admiral config info
- OVA deploys a VM to the target VC.
- Harbor and Admiral both run in this VM, on different ports:
  - Harbor https://_vm_address_:443
  - Admiral https://_vm_address_:8282
- Log ins for Harbor and Admiral will be separate in 1.1
- Harbor and Admiral UIs each have **Management** and **Registry** buttons, that in future releases will toggle between different views of a single, unified UI, but in 1.1 will send you to the login/UI of the other component.
- The plan to SSH into the deployed VM in order to run `vic-machine` to deploy/manage VCHs has been abandoned in favour of making `vic-machine` available for download from a Web server hosted on the VM, per vmware/vic#4254. You then run `vic-machine` on whatever system you like.
- The vSphere Client plug-in for VIC is also made available by the Web server.

stuclem:
VIC Engine download is at https://_vm_address_:9443

stuclem:
First wave of updates is https://github.com/vmware/vic-product/pull/123/commits/568479229bfcea61ca1beb46f5728f9fbb25caab

stuclem:
https://github.com/vmware/vic-product/issues/123 is merged. Further updates to be handled in other issues.

PR #123, opened 2017-03-29
Documenting OVA installer

stuclem:
Towards https://github.com/vmware/vic-product/issues/22 and https://github.com/vmware/vic-product/issues/49.


stuclem:
@frapposelli and @andrewtchin can I please ask you to take a look at the following:

- https://github.com/stuclem/vic-product/blob/ovainstaller/docs/user_doc/vic_vsphere_admin/installing_vic.md
- https://github.com/stuclem/vic-product/blob/ovainstaller/docs/user_doc/vic_vsphere_admin/deploy_vic_appliance.md

@jooskim can you please look at https://github.com/stuclem/vic-product/blob/ovainstaller/docs/user_doc/vic_vsphere_admin/plugins_vc_windows.md and https://github.com/stuclem/vic-product/blob/ovainstaller/docs/user_doc/vic_vsphere_admin/plugins_vcsa.md.

Thanks! Most of the other changes in this PR are minor. Please DO NOT review anything about using VIC with Harbor/Registry. These topics have been renamed, which is why they show up here, but are in need of a total rewrite.

docs/user_doc/vic_vsphere_admin/deploy_vic_appliance.md
```diff
@@ -0,0 +1,65 @@
...
+    - To use custom certificates to authenticate connections to vSphere Integrated Containers Management Portal, optionally paste the content of the appropriate certificate and key files in the **SSL Cert** and **SSL Cert Key** text boxes. Leave the text boxes blank to use auto-generated certificates.
+
+8. Click **Next** and **Finish** to deploy the vSphere Integrated Containers appliance.
+
+**Result**
+
+When the deployment completes, the appliance makes the vSphere Integrated Containers services available on the following ports:   
+
+- vSphere Integrated Containers Engine download: https://<i>vic_appliance_address</i>:9443
+- vSphere Integrated Containers client plug-in upload: https://<i>vic_appliance_address</i>:9443
```
andrewtchin: The main file that the user should be interested in here is the vic-engine-***.tar.gz which is the equivalent to what's currently hosted on Bintray. The other `.zip` files are more for internal use by the UI plugin installer scripts so the user doesn't need to download those. The UI plugin installer scripts are themselves included in the `tar.gz` bundle. So I think this second entry might not be needed.

docs/user_doc/vic_vsphere_admin/plugins_vcsa.md
```diff
@@ -0,0 +1,58 @@
...
+      <pre>if [[ $(echo $OS | grep -i "darwin") ]] ; then
+       PLUGIN_MANAGER_BIN="../../vic-ui-darwin"
+      else
+       PLUGIN_MANAGER_BIN="../../vic-ui-windows"</pre>
+
+7. Open a command prompt, navigate to `/vic/ui/VCSA`, and run the installer.
+   <pre>./install.sh</pre>
+
+    - Make sure that `install.sh` is executable by running `chmod` before you run it.
+    - On Windows systems, run `install.sh` in a UNIX shell that supports SSH and SCP, for example Cygwyn or Git Bash. Do not use Windows 10 native Bash.
```
jooskim: Since we will drop the no-webserver option it no longer requires SSH or SCP.

docs/user_doc/vic_vsphere_admin/plugins_vcsa.md
```diff
@@ -0,0 +1,58 @@
...
+  
+9. Enter the user name and password for the vCenter Server administrator account.
+
+10. Enter the `root` password for the vCenter Server Appliance.
```
jooskim: Assuming OVA is the only method for UI installation, which it should be, we don't need this. As soon as the installer script detects that a legit path to the plugin file it won't make any SSH connections.

andrewtchin: Not all users will install through the OVA, currently we are still going to distribute the standalone .tar.gz

jooskim: Yeah forgot to mention that. Then it's probably worth mentioning that users could use their own web server with HTTPS enabled, but as long as any sort of web server is used the current installer won't ask for root credentials.

docs/user_doc/vic_vsphere_admin/deploy_vic_appliance.md
stuclem: Thanks @andrewtchin. I am aware that the user won't need to download the zip for the UI, but we need to advertise its existence because users need to paste the URL into the UI installer script. However maybe it's misleading to give the URL here, as it will tempt people to download it.

docs/user_doc/vic_vsphere_admin/plugins_vcsa.md
stuclem: Ah, @jooskim for the VCSA deployment I thought that the no-webserver method was the easiest. That's what I have doc'd for VCSA, for both Flex and H5. Should I use the webserver method for both VC-on-Windows and VCSA?

docs/user_doc/vic_vsphere_admin/plugins_vcsa.md
stuclem: @andrewtchin and @jooskim, what I am documenting here is the official VMware release, rather than any open source releases. My understanding is that the only means of installing the official release will be the OVA. Will we still be shipping separate downloads for the 3 components with the official release?

stuclem: @jooskim, I'm a bit confused. For the installation on VCSA, I chose only to document the non-webserver method, which I believe is the more straightforward of the two. So, in that case, isn't the root password still required? My Nimbus VCSA has expired, so I need to deploy another one to test this out.

docs/user_doc/vic_vsphere_admin/deploy_vic_appliance.md
stuclem: Added that you don't need to download the plug-in files, but pass the URL to the plug-in installer script.

stuclem:
Thanks @frapposelli !

docs/user_doc/vic_vsphere_admin/plugins_vcsa.md
jooskim: Sorry for going back and forth, this really gets me confused sometimes. Like Andrew confirmed yesterday, we are not replacing one method with another but keep both methods in there. So please correct me if I am wrong. It seems that you decided to choose what looks the easiest way for users, but what I think is logical is to present two options equally - with and without webserver.

However, `no-webserver` method is inherently prone to cause human errors, and the way vSphere Client reacts when first accessed after deployment is slightly different between the two methods. I've confirmed yesterday that I had to restart the H5C web server to see the plugin when deploying without a webserver, whereas the web server method worked out of the box. Therefore I have to say I do not recommend the `no-webserver` method as the standard procedure for plugin deployment.

stuclem: Thanks @jooskim for the explanations, which clarify things a lot. In this case, maybe we should drop the no-webserver method from the official docs for both VCSA and Windows. Since we're documenting the official product, and since Karthik confirmed to me today that the only means of delivery of the official product is the OVA, everybody who has the official release has access to the web server in the VIC appliance VM. If deployment via Web server is more reliable too, then we should stick with just that method. The docs about how to deploy an unofficial build without a Web server have been written, so we can find a place to stash them that is outside of the official doc. What do you think?

docs/user_doc/vic_vsphere_admin/plugins_vcsa.md
jooskim: @stuclem That's a great point. I'm not sure how exactly the official VMware release will look. My thought regarding to your second comment is that unless we have the OVA ready in the tar ball, we keep the both methods documented as I posted a comment above. I don't think we do that as of now. @frapposelli could you please confirm if that's in the plan?

But yeah, assuming the `no-webserver` approach is used all my comments above could be ignored.

docs/user_doc/vic_vsphere_admin/plugins_vcsa.md
jooskim: GREAT plan. I like it very much

stuclem:
I have added the new OVA options for the file server. @frapposelli can you please take another look?

@jooskim I now use the Web server method for both VCSA and Windows VC. Can you also please take another look?

Thanks both!

stuclem:
Thanks @jooskim! Will merge this now...

PR #136, opened 2017-04-05
Removed quotes from configs file for Windows UI deployment.

stuclem:
Removed erroneous quotation marks from the `configs` file settings for deployment of the client plug-ins for vCenter Server on Windows.

@jooskim can you please confirm that you use quotes for VCSA but not for Windows?


stuclem:
Thanks @jooskim !

Issue #46, opened 2017-03-21
Document vic-machine upgrade --rollback


stuclem:
_From @stuclem on March 6, 2017 11:56_

Per https://github.com/vmware/vic/issues/3765

_Copied from original issue: vmware/vic#4143_

stuclem:
https://github.com/vmware/vic/issues/3765 is implemented in PR https://github.com/vmware/vic/pull/4122 (not merged yet)

stuclem:
PR https://github.com/vmware/vic/pull/4122 is now merged.

stuclem:
To document after 0.9

stuclem:
Added `--rollback` in the topic on upgrading VCHs and in the `vic-machine upgrade` options.

Change is in https://github.com/stuclem/vic-product/commit/d518e4060a281486d20455a7acf10e361ba3cd34.

PR #143, opened 2017-04-11
Added --rollback option.

stuclem:
Per title. Fixes https://github.com/vmware/vic-product/issues/46.

@gigawhitlocks can you please take a look? Thanks!

### DIFF
--- a/docs/user_doc/vic_vsphere_admin/upgrade_vch_options.md
+++ b/docs/user_doc/vic_vsphere_admin/upgrade_vch_options.md
@@ -46,6 +46,6 @@ If your vSphere environment uses untrusted, self-signed certificates, you can us
 
 Short name: None
 
-Rolls an upgraded VCH back to its previous version, for example if upgrade failed. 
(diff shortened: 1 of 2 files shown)


docs/user_doc/vic_vsphere_admin/upgrade_vch_options.md
```diff
@@ -41,3 +41,11 @@ Forces `vic-machine upgrade` to ignore warnings and continue with the upgrade of
 If your vSphere environment uses untrusted, self-signed certificates, you can use the `--force` option to upgrade a VCH without providing the thumbprint of the vCenter Server or ESXi host in the `thumbprint` option. 
 
 <pre>--force</pre>
+
+### `--rollback` ###
+
+Short name: None
+
+Rolls an upgraded VCH back to its previous version, for example if upgrade failed.
```
gigawhitlocks: I'm struggling with the wording myself, but I would suggest adding something along the lines of "--rollback only supports rolling back one version at this time."

gigawhitlocks:
One very small comment, otherwise LGTM

stuclem:
Thanks @gigawhitlocks. Added the following:

----

Rolls a VCH back to its previous version, for example if upgrade failed. Before starting the upgrade process, `vic-machine upgrade` takes a snapshot of the existing VCH. The upgrade process deletes older snapshots from any previous upgrades. The `--rollback` option reverts an upgraded VCH to the snapshot of the previous deployment. Because `vic-machine upgrade` only retains one snapshot, you can only use `--rollback` to revert the VCH to the version that immediately precedes the most recent upgrade.  

----

Simple enough change, so I think that it's safe to merge this.

Issue #45, opened 2017-03-21
Document ESX firewall configuration mechanism


stuclem:
_From @stuclem on February 21, 2017 15:39_

Per https://github.com/vmware/vic/issues/3979. 

_Copied from original issue: vmware/vic#3991_

stuclem:
See also https://github.com/vmware/vic/issues/3801, specifically about VIBs.

stuclem:
See also https://github.com/vmware/vic/issues/3927 and https://github.com/vmware/vic/issues/3643

stuclem:
Looks like https://github.com/vmware/vic/issues/3979 has priority over the VIB idea.

stuclem:
_From @andrewtchin on March 7, 2017 19:9_

```
### vic-machine update firewall
                                                                                                         
The `vic-machine update firewall` command can be used to modify the firewall on the ESXi host or the ESXi hosts in a cluster. 

With the flags `--allow` and `--deny` this tool enables and disables the ruleset "vSPC", respectively. When enabled, "vSPC" rule allows all outbound TCP traffic from the target host or hosts. If the rule is disabled, the firewall must be configured via another method to allow outbound 2377/tcp or VIC Engine will not function.

The firewall is not modified by `vic-machine create` or `vic-machine delete`. This command may be used before `vic-machine create` and after `vic-machine delete` to configure the firewall. 

To enable the rule, which permits all outbound TCP traffic from the host, use `--allow`:
<pre>vic-machine-<i>operating_system</i> update firewall
--target 'Administrator@vsphere.local':<i>password</i>@<i>vcenter_server_address</i>/dc1
--compute-resource cluster1
--thumbprint=<i>thumbprint</i>
--allow
</pre>

To disable the rule, use `--deny`:
<pre>vic-machine-<i>operating_system</i> update firewall
--target 'Administrator@vsphere.local':<i>password</i>@<i>vcenter_server_address</i>/dc1
--compute-resource cluster1
--thumbprint=<i>thumbprint</i>
--deny
</pre>
```

Also update https://github.com/vmware/vic/blob/master/doc/user_doc/vic_installation/ts_firewall_error.md

stuclem:
There is an ongoing discussion in https://github.com/vmware/vic/pull/4165 about whether this should be `vic-machine infra` or `vic-machine update`.

stuclem:
_From @andrewtchin on March 8, 2017 17:48_

Updated the above to reflect use of `vic-machine update`

[...]

PR #144, opened 2017-04-11
Added firewall update command

stuclem:
Per title. 

Fixes https://github.com/vmware/vic-product/issues/45.

@andrewtchin can you please take a look? Thanks!

### DIFF
--- a/docs/user_doc/vic_vsphere_admin/ts_firewall_error.md
+++ b/docs/user_doc/vic_vsphere_admin/ts_firewall_error.md
@@ -12,22 +12,4 @@ ESXi hosts communicate with the VCHs through port 2377 via Serial Over LAN. For 
 
 ## Solution ##
 
-The `vic-machine` utility includes an `update firewall` command, that you can use to modify the firewall on the ESXi host or the ESXi hosts in a cluster. 
-
-You use `--allow` and `--deny` flags to enable and disable a firewall rule named `vSPC`. When enabled, the `vSPC` rule allows all outbound TCP traffic from the target host or hosts. If you disable the rule, you must configure the firewall via another method to allow outbound connections on port 2377 over TCP. If you do not enable the rule or configure the firewall, vSphere Integrated Containers does not function, and you cannot deploy VCHs.
-
-Neither of the `vic-machine create` or `vic-machine delete` commands modify the firewall. You can run `vic-machine update firewall --allow` before you run `vic-machine create` and run `vic-machine update firewall --deny` after `vic-machine delete`. 
-
-<pre>vic-machine-windows update firewall 
---target 'Administrator@vsphere.local':'<i>password</i>'@<i>vcenter_server_address</i>/dc1 
---compute-resource cluster1
---thumbprint <i>thumbprint</i> 
---allow
-</pre>
-
-<pre>vic-machine-windows update firewall 
---target 'Administrator@vsphere.local':'<i>password</i>'@<i>vcenter_server_address</i>/dc1 
---compute-resource cluster1
---thumbprint <i>thumbprint</i> 
---deny
-</pre>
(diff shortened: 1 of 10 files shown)


docs/user_doc/vic_vsphere_admin/ts_firewall_error.md
```diff
@@ -12,37 +12,24 @@ ESXi hosts communicate with the VCHs through port 2377 via Serial Over LAN. For
...
-    &lt;protocol&gt;tcp&lt;/protocol&gt;
-    &lt;port type='dst'&gt;2377&lt;/port&gt;
-  &lt;/rule&gt;
-  &lt;enabled&gt;true&lt;/enabled&gt;
-  &lt;required&gt;true&lt;/required&gt;
-&lt;/service&gt;
-</pre>
+The `vic-machine` utility includes an `update firewall` command, that you can use to modify the firewall on the ESXi host or the ESXi hosts in a cluster. 
+
+You use `--allow` and `--deny` flags to enable and disable the `vSPC` ruleset. When enabled, the `vSPC` rule allows all outbound TCP traffic from the target host or hosts. If you disable the rule, you must configure the firewall via another method to allow outbound connections on port 2377 over TCP. If you do not enable the rule or configure the firewall, vSphere Integrated Containers does not function, and you cannot deploy VCHs.
```
andrewtchin: VCHes?

docs/user_doc/vic_vsphere_admin/ts_firewall_error.md
```diff
@@ -12,37 +12,24 @@ ESXi hosts communicate with the VCHs through port 2377 via Serial Over LAN. For
...
+<pre>vic-machine-windows update firewall 
+--target 'Administrator@vsphere.local':'<i>password</i>'@<i>vcenter_server_address</i>/dc1 
+--compute-resource cluster1
+--thumbprint <i>thumbprint</i> 
+--deny
+</pre>
 
-  ```$ esxcli network firewall set --enabled false``` 
\ No newline at end of file
+**NOTE**: Setting the `vSPC` rule does not persist through an ESXi host reboot. If you reboot an ESXi host on which you ran `vic-machine-windows update firewall` to enable the rule, you must run `vic-machine-windows update firewall` again to reenable the rule.
```
andrewtchin: On both my fusion ESXi 6.5 and @chengwang86's latest build, when using the `update firewall` command the rule persisted between restarts. I would also think persistence would be the expected behavior because this command is using the same API as the UI and I wouldn't expect the user to have to reconfigure those every reboot. Could you recheck this?

docs/user_doc/vic_vsphere_admin/ts_firewall_error.md
stuclem: Nope! Since VCH is already an abbreviation, there's no need to add the 'e'. Even if the plural of "sandwich" is "sandwiches" :)

docs/user_doc/vic_vsphere_admin/ts_firewall_error.md
stuclem: I had made a note in the issue that stated that the rule isn't persistent, but I just rebooted my ESX, saw that the `vSPC` rule was still set to `true`, and was able to deploy a VCH. So, I removed the note about it not persisting.

stuclem:
@andrewtchin I thought about this a bit and decided that this info needs to be in a topic of its own, right up front before you start deploying VCHs, rather than in a troubleshooting topic that tells you what you should have done before you tried to deploy a VCH and hit the firewall error. So I added the info in a new topic, and updated the troubleshooting topic to point to the new topic. All existing links to the TS topic now point to the new one. I think that this is better.

Can you please take another look?

andrewtchin:
lgtm

stuclem:
Thanks @andrewtchin !

Issue #140, opened 2017-04-07
Doc: Add the version in which support was added for Docker commands


stuclem:
Per request from @karthik-narayan.

See https://docs.google.com/spreadsheets/d/11yxTsO7Py248eQT430H48n3civ0CNalNKrr5pU8Hr1I/edit?ts=58e6d110#gid=0 for the version information.

stuclem:
Added versions in which support for commands was added in https://github.com/stuclem/vic-product/commit/13cfc01cff3010eb9961d2aabf75b6a49b820430.

Issue #139, opened 2017-04-07
Doc: Clarify that compose code sample is not complete


stuclem:
Per email from Dani Barkelew:

-----

Could the example in the document for the docker-compose-simple.yml file on step 4 at https://vmware.github.io/vic-product/assets/files/html/0.8/vic_app_dev/build_app_with_vic.html  get updated to have the complete example or warning that the blocked code is not a complete example?

Had a customer who did not think to click the link under the code block https://github.com/vmware/vic/tree/master/demos/compose/voting-app to get the actual compose yml needed with complete environment variables  to complete the demo walk through. When I went through the example, has I not already been in the troubleshooting mind set I might have also assumed the code block was complete for the demo. 

-----

stuclem:
Reworded to "The example below shows an **excerpt** of the YML file after the modifications:"

Change is in https://github.com/stuclem/vic-product/commit/543c379c7f9fb7dae3c884f15c2c906cf63c42c1

PR #159, opened 2017-04-15
Added versions in which Docker command support was added

stuclem:
Also clarified that the docker compose example is an excerpt, rather than the full example.

- Fixes https://github.com/vmware/vic-product/issues/140
- Fixes https://github.com/vmware/vic-product/issues/139


stuclem:
Simple changes so merging without need for review.

PR #169, opened 2017-04-17
It looks like non-root users are now supported.

stuclem:
While going through the list of known issues from 0.9 for the 1.1 release notes, I saw that https://github.com/vmware/vic/issues/1279 is now closed and non-root users are supported. 

@emlin can you please confirm?


emlin:
Yeah, it's fixed with same way as docker

stuclem:
Great, thanks @emlin !

PR #162, opened 2017-04-15
Added example of combining custom and autogen certs

stuclem:
Towards https://github.com/vmware/vic-product/issues/44

@andrewtchin and @hickeng can you please take a look?

PR #193, opened 2017-05-01
Comments from George about cert combos.

stuclem:
Following @hickeng 's comments in https://github.com/vmware/vic-product/pull/162#pullrequestreview-33089799, which fixes https://github.com/vmware/vic-product/issues/44.

### DIFF
--- a/docs/user_doc/vic_vsphere_admin/vch_installer_examples.md
+++ b/docs/user_doc/vic_vsphere_admin/vch_installer_examples.md
@@ -356,7 +356,7 @@ For more information about using custom server certificates, see the section [Re
 
 ### Combine Custom Server Certificates and Auto-Generated Client Certificates {#certcombo}
 
-You can create a VCH so that it uses a custom certificate to authenticate with the VCH Admin portal and an auto-generated certificate to authenticate with Docker clients. Specifying the paths to the `key.pem` and `cert.pem` files in the `--key` and `--cert` options is sufficient if the certificate is signed by a CA that is trusted by the system on which you run `vic-machine` and that is also trusted by the users who connect to the VCH. However, if the certificate is signed by a CA that is not trusted by the system, the installer cannot validate the certificate chain and `vic-machine create` fails with a certificate validation warning: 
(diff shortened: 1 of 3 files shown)


stuclem:
@hickeng can you please take another look?

docs/user_doc/vic_vsphere_admin/vch_installer_examples.md
```diff
@@ -356,7 +356,7 @@ For more information about using custom server certificates, see the section [Re
...
+You can create a VCH so that it uses a custom certificate to authenticate the VCH Admin portal and an auto-generated certificate to authenticate Docker clients. Specifying the paths to the `key.pem` and `cert.pem` files in the `--key` and `--cert` options is sufficient if the certificate is signed by a CA that is trusted by the system on which you run `vic-machine` and that is also trusted by the users who connect to the VCH. However, if the certificate is signed by a CA that is not trusted by the system, the installer cannot validate the certificate chain and `vic-machine create` fails with a certificate validation warning:
```
hickeng: > You can create a VCH so that it uses a custom certificate to authenticate the VCH Admin portal and an auto-generated certificate to authenticate Docker clients. Specifying the paths to the `key.pem` and `cert.pem` files in the `--key` and `--cert` options is sufficient

@stuclem I love concise documentation. I think we may be trying to be too concise. While the paragraph is correct I fear it will not be clear to people because we're mixing multiple things:
1. what a custom server certificate provides and when you may want to use it (ability of client to trust server, and how it presents in a browser)
1a. what requirements using a custom cert adds (adding an appropriate CA to the local certificate pool)
2. How to configure a VCH with custom server cert
3. Generation (or specification) of client certificates


The phrasing I was trying before concluding we're trying to mix too much - as you can see I've dropped the 1a part of it completely as I could not find a concise why of phasing it that I thought added clarity:
  * You can create a VCH with a custom server certificate by specifying the `-cert` and `-key` options, supplying paths to the custom `server-cert.pem` and `server-key.pem` files. The key should be un-encrypted. Generation of client certificates is not impacted by the `-cert` and `-key` options.

stuclem: Updated accordingly in https://github.com/vmware/vic-product/pull/193/commits/457f079522b8b7196abfbed7141a9c4815c710d9.

stuclem:
Integrated @hickeng 's comments from https://github.com/vmware/vic-product/pull/193#pullrequestreview-35601229 in https://github.com/vmware/vic-product/pull/193/commits/457f079522b8b7196abfbed7141a9c4815c710d9. Think this is safe to merge.

Issue #151, opened 2017-04-13
Document how to handle expired password


stuclem:
Per https://github.com/vmware/vic/issues/4214, which is implemented in PR https://github.com/vmware/vic/pull/4679. 

@hickeng confirms that this is in 1.1.

stuclem:
From @hickeng via Slack: 

You should see a message short of like:
```
Warning: your password will expire in 0 days
root@ [ ~ ]# date
Wed Apr 12 17:08:28 UTC 2017
```

when you log in

If the password has expired then you will no longer be able to log in by ssh key

If the password has expired an you try to log in interactively then you'll be prompted to change it (i.e. it doesn't lock the account)

If you rerun --enableSSH without specifying --rootpw then the password expiration will be set to 1 day in the future (using calendar maths, so around midnight that could be a very short period - I've not played with this to find out exact behaviour) but the current password value will be preserved - i.e. just updates expiration (edited)
 
If a longer expiration is required then it _should_ be possible to change it via `passwd` once you're logged in and get a 90 day expiry set - (unconfirmed)

stuclem:
Attempted to add the above to the topics on debug. Change is in https://github.com/stuclem/vic-product/commit/a8a9cab01b8c18212a2772901d716cec9e209166.

stuclem:
Changes merged in https://github.com/vmware/vic-product/issues/158, but leaving this open for final verification by @hickeng.

hickeng:
@stuclem I've added a comment to the PR - it's correct, but I thought I'd expand on one point a little. Up to you if you want to add it to the docs.

PR #158, opened 2017-04-15
Updates related to debug and logging

stuclem:
- Towards https://github.com/vmware/vic-product/issues/68 (`vic-machine create debug` levels)
- Towards https://github.com/vmware/vic-product/issues/151 (`vic-machine debug` password expiry)
- Towards https://github.com/vmware/vic-product/issues/55 (log rotation)
- Towards https://github.com/vmware/vic-product/issues/52 (added links in log bundle topic)

@hickeng can you please take a look at the two debug issues?

@fdawg4l can you please look at the logging issues? 

Thanks!

PR #206, opened 2017-05-03
Correction in password expiry doc

stuclem:
Fixes https://github.com/vmware/vic-product/issues/151, following comment from @hickeng in PR https://github.com/vmware/vic-product/pull/158. 

@hickeng is this OK now?

### DIFF
--- a/docs/user_doc/vic_vsphere_admin/debug_vch_options.md
+++ b/docs/user_doc/vic_vsphere_admin/debug_vch_options.md
@@ -14,7 +14,7 @@ Short name: `--pw`
 
 Set a new password for the root user account on the VCH endpoint VM. Setting a password on the VCH allows you to access the VCH by using the VM console. If you also set the `--enable-ssh` option, you can use this password to connect to the VCH by using SSH. 
 
-When you use the password to log in to a VCH, you see the message that the password will expire in 0 days. To obtain a longer expiration period, use the Linux `passwd` command in the endpoint VM to set a new password. If the password expires, the VCH does not revert to the default security configuration from before you ran `vic-machine debug`. If you attempt to log in after the password has expired, you see a prompt to change the password, and you can then log in. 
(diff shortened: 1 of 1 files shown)


docs/user_doc/vic_vsphere_admin/debug_vch_options.md
```diff
@@ -14,7 +14,7 @@ Short name: `--pw`
 
 Set a new password for the root user account on the VCH endpoint VM. Setting a password on the VCH allows you to access the VCH by using the VM console. If you also set the `--enable-ssh` option, you can use this password to connect to the VCH by using SSH. 
 
-When you use the password to log in to a VCH, you see the message that the password will expire in 0 days. To obtain a longer expiration period, use the Linux `passwd` command in the endpoint VM to set a new password. If you attempt to log in after the password has expired, you see a prompt to change the password, in which case you must run `vic-machine debug --rootpw` again. 
+When you use the password to log in to a VCH, you see the message that the password will expire in 0 days. To obtain a longer expiration period, use the Linux `passwd` command in the endpoint VM to set a new password. If the password expires, the VCH does not revert to the default security configuration from before you ran `vic-machine debug`. If you attempt to log in after the password has expired, you see a prompt to change the password, and you can then log in.
```
hickeng: > If you attempt to log in after the ....
If you attempt to log in using interactive password via terminal or ssh, you see a prompt to change the password. If you are using an SSH key then login will be denied until either the password is changed or `vic-machine debug` is run again.

@stuclem you don't actually need to provide a new password when re-running vic-machine - that was deliberate so that you could keep a stable password. This behaviour will become more sane once we have the account locked rather than just password expiry.

stuclem: Thanks @hickeng. Took your text above pretty much verbatim, so I think that we can merge this one.

Issue #204, opened 2017-05-03
Document that special characters are permitted in user passwords


stuclem:
I just created a user with a password that includes $ and !, and it worked. @reasonerjt can you please confirm that special characters are now permitted in passwords? Thanks!

reasonerjt:
It is, there was a bug some special chars are not allowed, and it's fixed in 1.1.1

stuclem:
OK, thanks @reasonerjt. Fix is in https://github.com/vmware/vic-product/issues/216.

PR #216, opened 2017-05-10
Added that passwords can include special characters.

stuclem:
Fixes https://github.com/vmware/vic-product/issues/204. 

@reasonerjt is this OK now? Thanks!


stuclem:
Such a simple change that I think it's safe to merge this.

Issue #181, opened 2017-04-26
Update volumes doc to indicate fstype support


stuclem:
From Slack:

```
douggie [7:01 PM] 

just had a question about the fstype on the volume create command.

douggie [7:02 PM] 
added this Plain Text snippet
docker -H vichost:2376 --tls volume create --driver=vsphere --name=test2 -o diskformat=zeroedthick -o size=500mb -o attach-as=independent_persistent -o fstype=ext2
Add Comment

douggie [7:03 PM] 
however when this is mount it is mounted as ext4 (edited)

douggie [7:04 PM] 
added this Plain Text snippet
root@8643967b93d0:~# blkid
/dev/sda: LABEL="containerfs" UUID="a4821c1c-f249-4b5e-95e2-1365d324ae59" TYPE="ext4" 
/dev/sr0: LABEL="ISOIMAGE" TYPE="iso9660" 
/dev/sdb: LABEL="ad0234829205b903" UUID="930b9502-4557-4ded-99f2-9567009a1e3d" TYPE="ext4" 
Add Comment

matthewavery [7:05 PM] 
@douggie It depends on the driver and volume store you are using. This looks like you are using the docker-vsphere-volume driver options? VIC does not support the docker-vsphere-volume driver. Since you have not specified a volume store here it means that you are targetting the `default` volume store. since it succeeds I assume you have one setup. for VIC( atleast currently) the fs type will depend on the volume store you target. if it is a vsphere datastore target then the object type will be a disk and we always format it with `ext4` for now. (edited)

douggie [7:06 PM] 
we are using luks encryption and it is recommended to use a non-journaled file system as the only cipher that works with the vsphere is the one built into the linux kernel

[7:07] 
cbc

matthewavery [7:09 PM] 
We do not currently support other file system types at volume create time other than ext4 and nfs(coming soon). You can make a ticket for additional file system support and we will triage it :slightly_smiling_face: then we can hold a more indepth discussion there that will involve engineers from the different components.

douggie [7:11 PM] 
sure, so for now only ext4 is supported, so why have the -o fstype= option?

matthewavery [7:35 PM] 
We do not :slightly_smiling_face: I believe that is a docker-vsphere-volume option. Any option that we do not recognize is ignored.

[7:36] 
currently the options we support are `-o Capacity` and `-o VolumeStore`

matthewavery [8:08 PM] 
[...]

PR #229, opened 2017-05-17
Added that VIC only supports EXT4 volumes.

stuclem:
Fixes https://github.com/vmware/vic-product/issues/181. @matthewavery do we need add more? Thanks!


stuclem:
@matthewavery yep, but since this update is going to go out with 1.1.1, we shouldn't mention NFS just yet. Thanks!

PR #244, opened 2017-05-19
Added instruction to use colon-delimitation in thumbprint everywhere.

stuclem:
Per title, following comment from @alextopuzov that this was not mentioned consistently.


alextopuzov:
LGTM.

PR #255, opened 2017-05-29
Adding an Edit link to all pages in HTML output.

stuclem:
Installed the gitbook edit-link plug-in, that adds an EDIT THIS PAGE link directly at the top of every page of the HTML output:

editpage

Clicking this link takes you directly to the source file for that page in the Github repo.

However, the `vic-product` repo is unprotected, so it is possible to click the Edit button and commit edits directly into `master `, which we do not want. So, before I merge this PR, @frapposelli can we configure the  `vic-product` repo so that `master` is protected, and any edits have to be submitted via a branch/PR?

To compare, here's what you see when you edit a page in GitHub in `vic-product`:

commitmaster

And here's what you see when you do the same in GitHub in  `vic`:

nocommitmaster


stuclem:
@frapposelli has now made `vic-product:master` and `vic-product:gh-pages` into protected branches, so it's now safe to merge this one. 

@alextopuzov can you please review this PR? Thanks!

PR #272, opened 2017-06-07
Adding new network graphic from UX.

stuclem:
This version is a bit small and fuzzy, but it's still a lot better than my effort.


stuclem:
@lweitzman do we have a higher-res version? 
Thanks!

stuclem:
Received newer version from @ali5ter. Still not the final version, but it's good enough to publish. I will update it once we get the very final incarnation, but in the meantime I want to push this live.

Issue #275, opened 2017-06-09
Document how to expand OVA storage for Registry


stuclem:
Per https://vmware-com.socialcast.com/messages/35616151?sso_login=true, we need to document how to expand the storage for Registry.

stuclem:
Changes are in https://github.com/vmware/vic-product/pull/345/commits/7c4da17a2c7060ef9f5208cb478f36fdde1b74d9.

PR #345, opened 2017-06-16
Added topic on reconfiguring appliance and content reorg

stuclem:
Fixes https://github.com/vmware/vic-product/issues/275 and does a bit of content reorg.

@andrewtchin and @frapposelli can you please take a look?


stuclem:
Thanks @andrewtchin !

Issue #247, opened 2017-05-23
Document that container-network needs network name as well as port group name


stuclem:
Per https://github.com/vmware/vic/issues/5066.

Applies to 1.1.x docs as well as 1.2.

morris-jason:
More specifically, a network name is mandatory if the port group name contains a space. Otherwise, the network name is optional. In both cases the network name cannot contain a space.

stuclem:
Fix is in https://github.com/vmware/vic-product/pull/346/commits/3529b828a74216f9326a0a5b747b08dcb3fc38fe.

PR #346, opened 2017-06-16
Added warning about spaces in container network port group names.

stuclem:
Fixes https://github.com/vmware/vic-product/issues/247. 

@morris-jason can you please take a look?


stuclem:
Thanks @morris-jason !

PR #348, opened 2017-06-16
Added "Resources" to paths for rps in clusters.

stuclem:
Per customer comment via Slack, I reinserted "Resources" into the paths when specifying resource pools in clusters in `--compute-resource`. 

@emlin is this correct? Thanks!

### DIFF
--- a/docs/user_doc/vic_vsphere_admin/vch_installer_examples.md
+++ b/docs/user_doc/vic_vsphere_admin/vch_installer_examples.md
@@ -100,16 +100,16 @@ This example deploys a VCH with the following configuration:
 
 ### Deploy to a Resource Pool in a vCenter Server Cluster {#rp_cluster}
 
-To deploy a VCH in a resource pool in a vCenter Server cluster, you specify the names of the cluster and resource pool in the `compute-resource` option.
+To deploy a VCH in a resource pool in a vCenter Server cluster, you specify the resource pool in the `compute-resource` option.
 
 This example deploys a VCH with the following configuration:
 
 - Specifies the user name, password, datacenter, image store, bridge network, and name for the VCH.
-- Designates `rp 1` in cluster `cluster 1` as the resource pool in which to place the VCH. Note that the resource pool and cluster names are wrapped in quotes, because they contain spaces. Use single quotes if you are using `vic-machine` on a Linux or Mac OS system and double quotes on a Windows system.
+- Designates `rp 1` as the resource pool in which to place the VCH. In this example, the resource pool name `rp 1` is unique across all hosts and clusters, so you only need to specify the resource pool name.
 
 <pre>vic-machine-<i>operating_system</i> create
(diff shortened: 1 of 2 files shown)


emlin:
@stuclem my understanding is that we don't need to have Resources inserted between cluster and resource pool name.
Is there an issue filed for the problem you mentioned from slack?

stuclem:
@emlin there isn't an issue for this. I have just re-tested various setups with resource pools and conclude the following:

- If the resource pool name is unique in VC, you only need to specify the resource pool name, even if there are multiple hosts and clusters.
- If different hosts or clusters contain resource pools that have the same name, e.g. `cluster1` includes `rp1` and `cluster2` also has an RP called `rp1`, you must specify the full path to the resource pool.

So, the customer was correct in that `--compute-resource cluster/rp_name` is wrong, but if the resource pool name is not unique, then you do need the full path, including `Resources`, e.g. `--compute-resource cluster2/Resources/rp1`. I have updated the options reference and the examples accordingly in the latest commits.

emlin:
@stuclem thanks for verify it.
@dougm Can you help to confirm if we change the behavior to this way, or it's a bug so far? I tried the create, it does the way as @stuclem said, and it's different with our original design as <cluster name>/<rp name>. thanks

dougm:
Right, if the resource pool name is unique, no need to specify the path.  Likewise for other resource types.
@emlin are you saying we should always require a path to the resource pool?

stuclem:
Hi @dougm, no, @emlin was questioning the fact that I had restored the full paths to RPs in the docs, based on customer feedback. Doc previously stated that you specify `cluster_name/rp_name`, which was incorrect. Customer requested I update it to  `cluster_name/Resources/rp_name`, which worked for him. Then I re-tested it and worked out that you only need the full path if RP name is not unique.

emlin:
@dougm previously, user can specify `cluster name/rp name`, but now it becomes `cluster name/Resources/rp name`, is that by design? Just want to make sure that's not some bug we'll fix in the future

stuclem:
Just to clarify, I'm talking about 1.1 here, not 1.2. I tested:

- `--compute-resource cluster_name/Resources/rp_name` works in all cases
- `--compute-resource rp_name` works if `rp_name` is unique across all hosts/clusters
- `--compute-resource cluster_name/rp_name` didn't work in any case

emlin:
thanks @stuclem for clarification. Looks like my mind is not refreshed for some time. Then I'm fine with it.

dougm:
ok, I don't recall `cluster_name/rp_name` without the parent pool ("Resources") in the path having worked
in the past.  We may have been doing something special to support that, but personally I think the correct inventory is best as the same path works with other tools (such as govc).  And 1.2 behaves the same as 1.1 in this respect.

stuclem:
Great, thanks @emlin and @dougm !

stuclem:
♻️

stuclem:
♻️

PR #351, opened 2017-06-19
Latest version of networking graphic.

stuclem:


stuclem:
Thanks @alextopuzov !

PR #393, opened 2017-06-26
refactored engine installer and updated README requiring user login to webserver

jialin-li:


jialin-li:
can't get update to show up, will close this one and reopen another pr.

PR #399, opened 2017-06-28
Change kovd cmd option from --tls-certificate-key to --tls-key

maplain:
1. because of the change in goswagger


maplain:
moved to https://github.com/vmware/vic-product/pull/400

PR #403, opened 2017-06-28
move initialization services to fileserver

jialin-li:


installer/fileserver/server.go
```diff
@@ -81,17 +100,82 @@ func Init(conf *config) {
...
 	t.Certificates = []tls.Certificate{c.cert}
 	s := &http.Server{
 		Addr:      c.addr,
-		Handler:   http.FileServer(http.Dir(c.serveDir)),
+		Handler:   mux,
 		TLSConfig: t,
 	}
 
-	log.Infof("Starting server on %s", s.Addr)
+	log.Infof("Starting installer-engine server on %s", s.Addr)
```
morris-jason: Fileserver?

morris-jason: Fileserver?

installer/fileserver/html/exec.html
```diff
@@ -0,0 +1,77 @@
...
+        {{ end }}
+
+    </main>
+</div>
+</body>
+</html>
+
+
+
+
```
morris-jason: Not a fan of calling this file exec.html.

installer/fileserver/server.go
jialin-li: Oops, fixed!

PR #410, opened 2017-07-04
Add systemd path units for SSO integration

andrewtchin:
Start Admiral and Harbor after token file appears at `/etc/vmware/psc/admiral/token` and `/etc/vmware/psc/harbor/token`

PR #415, opened 2017-07-06
Run PSC register command during OVA init

anchal-agrawal:
This change runs the PSC `register` command during the OVA init step to generate the PSC-related files in `/etc/vmware/psc`. Issue vmware/vic#5551

Tested on top of https://github.com/vmware/vic-product/pull/410 - Admiral and Harbor start correctly after the token config file has been generated.

### DIFF
--- a/installer/fileserver/tasks.go
+++ b/installer/fileserver/tasks.go
@@ -1,0 +1,103 @@
+// Copyright 2017 VMware, Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+import (
+	"context"
+	"fmt"
+	"os/exec"
+	"strings"
+
+	log "github.com/sirupsen/logrus"
+
+	"github.com/vmware/vic-product/installer/lib"
+	"github.com/vmware/vic-product/installer/pkg/ip"
+	"github.com/vmware/vic/pkg/vsphere/optmanager"
+)
+
+const (
+	pscBinaryPath    = "/etc/vmware/admiral/admiral-auth-psc-1.2.0-SNAPSHOT-command.jar"
+	vcHostnameOption = "config.vpxd.hostnameUrl"
+	pscConfDir       = "/etc/vmware/psc"
+)
+
+// registerWithPSC runs the PSC register command to register VIC services with
+// the platforms services controller. The command generates a config file and
+// certificates to use while getting and renewing tokens.
+func registerWithPSC(ctx context.Context) error {
+	var err error
+
+	session := admin.Validator.Session
+	version := session.ServiceContent.About.Version
+	versionFields := strings.Split(version, ".")
+	if len(versionFields) > 2 {
+		// Set version in required format (x.y)
+		version = versionFields[0] + "." + versionFields[1]
+	}
+
+	// Obtain the admin user's domain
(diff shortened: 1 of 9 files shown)


installer/fileserver/tasks.go
```diff
@@ -0,0 +1,86 @@
...
+		return err
+	}
+
+	// Obtain the OVA VM's IP
+	vmIP, err := ip.InterfaceIPv4("eth0")
+	if err != nil {
+		return err
+	}
+
+	log.Infof("version: %s", version)
```
anchal-agrawal: These logs are for debugging only; I'll remove them before merging.

installer/fileserver/tasks.go
```diff
@@ -0,0 +1,86 @@
...
+		"-jar",
+		"/etc/vmware/admiral/admiral-auth-psc-1.2.0-SNAPSHOT-command.jar",
+		"--command=register",
+		"--clientName=admiral",
+		"--version=" + version,
+		"--tenant=" + domain,
+		"--domainController=" + vcHostName,
+		"--username=" + admin.User,
+		"--password=" + admin.Password,
+		"--admiralUrl=https://" + vmIP.String() + ":8282",
```
hmahmood: This is now in guestinfo at: `guestinfo.vicova.admiral.endpoint`.

installer/pkg/ip/ip.go
```diff
@@ -0,0 +1,48 @@
...
+// limitations under the License.
+
+package ip
+
+import (
+	"fmt"
+	"net"
+)
+
+// InterfaceIPv4 returns the IPv4 address of the input interface.
```
hmahmood: Returns the first ipv4 address only right? There may be multiple ipv4 addresses on the interface.

installer/fileserver/tasks.go
```diff
@@ -0,0 +1,86 @@
...
+
+	log.Infof("version: %s", version)
+	log.Infof("tenant: %s", domain)
+	log.Infof("vcHostName: %s", vcHostName)
+	log.Infof("vmIP: %s", vmIP.String())
+
+	cmdName := "java"
+	cmdArgs := []string{
+		"-jar",
+		"/etc/vmware/admiral/admiral-auth-psc-1.2.0-SNAPSHOT-command.jar",
```
hmahmood: This should be a `const`.

installer/fileserver/tasks.go
```diff
@@ -0,0 +1,86 @@
...
+		"--command=register",
+		"--clientName=admiral",
+		"--version=" + version,
+		"--tenant=" + domain,
+		"--domainController=" + vcHostName,
+		"--username=" + admin.User,
+		"--password=" + admin.Password,
+		"--admiralUrl=https://" + vmIP.String() + ":8282",
+		"--configDir=/etc/vmware/psc",
```
hmahmood: `const`

installer/fileserver/tasks.go
```diff
@@ -0,0 +1,86 @@
...
+	}
+
+	// Obtain the hostname of the vCenter host
+	vcHostName, err := optmanager.QueryOptionValue(ctx, session, "config.vpxd.hostnameUrl")
+	if err != nil {
+		return err
+	}
+
+	// Obtain the OVA VM's IP
+	vmIP, err := ip.InterfaceIPv4("eth0")
```
hmahmood: `eth0` should be `const`.

installer/fileserver/tasks.go
hmahmood: Does this directory exist when this command runs?

anchal-agrawal: It does: https://github.com/vmware/vic-product/blob/master/installer/packer/scripts/package_provisioning.sh#L32. Also, the PSC command creates the directory if it doesn't exist.

installer/pkg/ip/ip.go
anchal-agrawal: Correct. Updated func name to `FirstIPv4`.

installer/fileserver/tasks.go
anchal-agrawal: Since this code runs before that key is set in guestinfo, I'll keep this as-is. Working on fetching the Admiral port through ovfenv.

PR #417, opened 2017-07-06
Fixes #416: Improve clarity and correctness of app-dev networking, compose support, dockerfile support

### LINKED ISSUE #416: The section on VIC networking for app devs needs improving
While working on another issue, I tried to use our own documentation to help me figure out the networking. This led me to conclude that there were a few things about this particular page that could be significantly improved. Main points:

1) The section showing the use of multiple --net arguments for multiple bridge networks is wrong. This has never been the correct syntax.

2) There's not enough discussion of *why* you might choose one strategy over another. In particular, the question of why you might want to use a container network over port mapping is not discussed at all.

3) The connection between different types of network and how those manifest in the guest is useful to know when debugging.

4) The fact that you can't add a bridge network to a running container is important divergent behavior. In particular because we don't show an error message if you try to do it.

5) The section on External Networks was very confusing. When we say "external" networks, do we mean container networks? It seems like we do, but they show up as "bridge" driver in the output. It shows a vic-production network and a default-external, but default-external doesn't actually appear to be used in the example. I suspect this is very old content from the 0.8 days. I don't think it needs the network inspect output and I think we can replace entirely.

6) The section on multiple container networks is even more confusing. You can't connect a container to multiple container networks, for a start. It seems to be there as a means of explaining what a container network is, but that makes it seem distinct from an external network. Again, I think we can remove this pretty much entirely.

corrieb:
This may not be perfect, but it at least addresses the issues listed. We may want to link to a more detailed discussion on network topologies and isolation. Once we have DinV, we will also need to add  a section on what this means for networking.


stuclem:
@corrieb, I suggest that we merge this so that I can do an editorial pass. I'll then put it back to you in another PR for final approval. Does that sound good to you? 

Thank you VERY much for doing this!

corrieb:
I didn't mean to put all of these changes in the same PR, but then I forgot to create a branch, so you're getting all of my changes at once.

This is as much as I am able to do until I go on vacation next Tuesday. I'm perfectly happy for you to make whatever formatting or grammatical changes you wish.

The most important thing is that this is a vast improvement in terms of presenting customers with the means to get the best out of VIC. It is also a substantial upgrade in terms of getting to grips with Compose.

stuclem:
Merging this into master now. Will do an editorial pass over this in another PR.

PR #418, opened 2017-07-07
Add tlsverify functionality to DinV images

frapposelli:
This closes vmware/vic#4860


dinv/dinv-1.12/certs.go
```diff
@@ -0,0 +1,118 @@
...
+	for _, file := range files {
+		if _, err := os.Stat(file); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func getFirstIP(ifname string) (net.IP, error) {
```
andrewtchin: Can this be used? https://github.com/vmware/vic-product/blob/master/installer/pkg/ip/ip.go

dinv/dinv-1.12/certs.go
```diff
@@ -0,0 +1,118 @@
...
+	}
+
+	return nil, fmt.Errorf("Can't find %s IP address", ifname)
+}
+
+// If tlsverify is set, but no certificates are available, we create CA, server and
+// client certs, this is a sort of "last resort" for people who want to try out
+// dinv but not wanting to create certs on their machine, client certs can be
+// downloaded from /certs.
+func generateCACerts(ip net.IP) error {
```
frapposelli: the certs are only saved inside the dinv container or on a docker volume mounted on `/certs`

dinv/dinv-1.12/certs.go
frapposelli: interesting, it would be a drop in replacement but will also mean I have to pull down `github.com/vmware/vic-product` at build time (which is due to the peculiar dinv build process).

Right now I'll prefer to just keep it that way, as it's just a simple function, eventually we could move most of this code to the ova repo and have just a simple main package in the dinv directory.

PR #429, opened 2017-07-11
Create PSC tokens during OVA init

anchal-agrawal:
Fixes vmware/vic#5551

### DIFF
--- a/installer/packer/scripts/psc/put_token_guestinfo.sh
+++ b/installer/packer/scripts/psc/put_token_guestinfo.sh
@@ -16,5 +16,5 @@
 # This script grabs the engine PSC token and puts it in guestinfo.
 set -euf -o pipefail
 
-engine_psc_token=$(grep "access_token" /etc/vmware/psc/engine/tokens.properties | awk -F'=' '{print $2}')
+engine_psc_token=$(grep "access_token" /etc/vmware/psc/engine/tokens.properties | awk -F'access_token=' '{print $2}')
 /etc/vmware/set_guestinfo.sh "engine.token" $engine_psc_token
--- a/installer/packer/scripts/psc/put_token_guestinfo.sh
+++ b/installer/packer/scripts/psc/put_token_guestinfo.sh
@@ -1,0 +1,20 @@
+#!/usr/bin/bash
+# Copyright 2017 VMware, Inc. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# This script grabs the engine PSC token and puts it in guestinfo.
+set -euf -o pipefail
+
+engine_psc_token=$(grep "access_token" /etc/vmware/psc/engine/tokens.properties | awk -F'=' '{print $2}')
+/etc/vmware/set_guestinfo.sh "engine.token" $engine_psc_token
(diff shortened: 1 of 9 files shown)


installer/packer/scripts/systemd/psc/get_token.service
```diff
@@ -0,0 +1,13 @@
+[Unit]
+Description=OVA PSC Token Creation
+Documentation=http://github.com/vmware/vic-product/installer
+After=docker.service systemd-networkd.service systemd-resolved.service iptables.service ova-firewall.service admiral_startup.path
```
anchal-agrawal: @andrewtchin looking for feedback here, thanks!

installer/packer/scripts/psc/put_token_guestinfo.sh
```diff
@@ -0,0 +1,20 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# This script grabs the engine PSC token and puts it in guestinfo.
+set -euf -o pipefail
+
+engine_psc_token=$(grep "access_token" /etc/vmware/psc/engine/tokens.properties | awk -F'=' '{print $2}')
```
hmahmood: You could just merge this into `get_token.sh`, and get rid of the extra systemd unit.

anchal-agrawal: Is there a better way to do this? Specially if `=` may be present in the token itself.

anchal-agrawal: Having a separate script for putting the token in guestinfo would be useful while refreshing tokens. Thoughts?

anchal-agrawal: Fixed!

installer/packer/scripts/systemd/psc/get_token.service
andrewtchin: Does the admiral service need to be up all the way before you run get tokens? If so, it should be `admiral.service`
I think you can probably remove all the other units from `After` besides the correct admiral one

installer/packer/scripts/systemd/psc/get_token.service
```diff
@@ -0,0 +1,13 @@
...
+Before=harbor_startup.service
+
+[Service]
+Type=oneshot
+ExecStartPre=/usr/bin/chmod +x /etc/vmware/psc/get_token.sh
```
andrewtchin: Is `get_token.sh` checked in with `x` perms? Did it not propagate when copied by Packer?

installer/packer/scripts/system_settings.sh
```diff
@@ -31,6 +31,7 @@ systemctl enable ovf-network.service
 systemctl enable ova-firewall.service
 systemctl enable harbor_startup.path
 systemctl enable admiral_startup.path
+systemctl enable get_token.service
```
andrewtchin: If we need to wait for the user to enter credentials before running this unit, you can remove this line.

andrewtchin:
Where does the getting the token process use the admin credentials from the landing page?

anchal-agrawal:
> Where does the getting the token process use the admin credentials from the landing page?

@andrewtchin Only the `register` command needs user creds. The `get-token` command uses the config file generated by the `register` command.

installer/packer/scripts/system_settings.sh
anchal-agrawal: Good call - `get_token.service` doesn't need to start at boot time. It's started explicitly by `admiral_startup.service`.

installer/packer/scripts/systemd/psc/get_token.service
anchal-agrawal: It doesn't have the `x` bit set locally - I'll update it.

installer/packer/scripts/systemd/psc/get_token.service
anchal-agrawal: Re: ordering of getting tokens and admiral startup - the details aren't final yet, but for now the plan is to start admiral after the PSC config file  is present (`admiral_startup.path`).

I'll remove the other `After` deps.

PR #443, opened 2017-07-19
Add systemd timer to refresh PSC tokens

anchal-agrawal:
Fixes vmware/vic#5554

### DIFF
--- a/installer/packer/scripts/systemd/psc/get_token.timer
+++ b/installer/packer/scripts/systemd/psc/get_token.timer
@@ -3,7 +3,6 @@ Description=Run the PSC Get Token service every hour
 Documentation=http://github.com/vmware/vic-product/installer
 
 [Timer]
-OnActiveSec=1h
 OnUnitActiveSec=1h
 
 [Install]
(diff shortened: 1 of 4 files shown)


installer/packer/scripts/systemd/psc/get_token.timer
```diff
@@ -0,0 +1,10 @@
+[Unit]
+Description=Run the PSC Get Token service every hour
+Documentation=http://github.com/vmware/vic-product/installer
+
+[Timer]
+OnActiveSec=1h
```
anchal-agrawal: Optimization: removing `OnActiveSec` so `get_token.service` isn't activated 1 hour after the timer is activated, but rather, just at 1 hour intervals (`OnUnitActiveSec`) _after_ `get_token.service` has been activated once in `admiral_startup.service`. This helps the case where the user may not have entered creds to trigger `admiral_startup.service` within 1 hour of OVA startup.

This is an optimization because if `get_token.service` is triggered before the startup of Admiral and Harbor, the `psc-config.properties` file will not be present - `get_token.service` will fail and be a no-op.

PR #449, opened 2017-07-21
PSC: Use SAML auth for vSphere 6.0 and 6.5

anchal-agrawal:
This change sets the version parameter for the PSC jar to 6.0
regardless of the actual version, since for 1.2 we intend to use
only SAML (originally only for 6.0) for 6.5 as well.


installer/pkg/version/version.go
```diff
@@ -43,7 +43,7 @@ type Build struct {
 }
 
 func init() {
-	flag.BoolVar(&v, "version", false, "Show version info")
+	flag.BoolVar(&v, "ver", false, "Show version info")
```
anchal-agrawal: This change isn't part of the pull request - it fixes a panic in `vic-ova-ui` in the current master branch which causes the VM UI to not load on boot:
```
bin/vic-ova-ui flag redefined: version
panic: bin/vic-ova-ui flag redefined: version

goroutine 1 [running]:
flag.(*FlagSet).Var(0xc4200607e0, 0x1467960, 0x14b9489, 0xe13974, 0x7, 0xe1bea0, 0x11)
    /usr/local/go/src/flag/flag.go:793 +0x420
flag.BoolVar(0x14b9489, 0xe13974, 0x7, 0xc4201c2c00, 0xe1bea0, 0x11)
    /usr/local/go/src/flag/flag.go:572 +0x72
github.com/vmware/vic-product/installer/pkg/version.init.1()
    /home/anchala/go/src/github.com/vmware/vic-product/installer/pkg/version/version.go:46 +0x5c
github.com/vmware/vic-product/installer/pkg/version.init()
    /home/anchala/go/src/github.com/vmware/vic-product/installer/pkg/version/version.go:135 +0x5d
main.init()
    /home/anchala/go/src/github.com/vmware/vic-product/installer/ovatools/vic-ova-ui/network_status.go:83 +0x71
```

This seems to be a regression from https://github.com/vmware/vic-product/pull/448.

hmahmood: Do we know where the conflict is?

anchal-agrawal: I think this is because both `vic` and `vic-product` now have `pkg/version`, which in turn have their own `version` flags: https://github.com/vmware/vic/blob/master/pkg/version/version.go#L56

Since both these pkgs are present in `$GOPATH`, it's likely this is causing the `flag redefined: version` issue.

frapposelli: remove the `init()` function altogether, we will add it back once we align all the tools to the same cli library.

PR #458, opened 2017-07-25
Fixes certificate generation in dinv

frapposelli:


dinv/dinv-1.12/main.go
```diff
@@ -103,9 +113,14 @@ func main() {
 
 			if err := certsExist(files); err != nil {
 				log.Debug("Certs not available, generating...")
-				ip, err := getFirstIP("eth0")
-				if err != nil {
-					log.Fatal(err.Error())
+				var ip net.IP
```
anchal-agrawal: Can this block be extracted into a function? It's being used in several places.

frapposelli: there is no sharing between the two dinv version, this is being used only twice in the same file, at some point I'll break down all this stuff into a shared package, but not today :)

anchal-agrawal: Sounds good. I added https://github.com/vmware/vic-product/tree/master/installer/pkg/ip recently - we can enhance that pkg.

PR #465, opened 2017-07-26
OVA: start Admiral container with PSC parameters

anchal-agrawal:
This change modifies the docker run command for the Admiral
container to add parameters for PSC. All config/cert files
needed by Admiral are now present in one dir, which is
mounted as a volume in the container.

### DIFF
--- a/installer/packer/scripts/admiral/configure_admiral.sh
+++ b/installer/packer/scripts/admiral/configure_admiral.sh
@@ -174,10 +174,8 @@ $script_dir/set_guestinfo.sh admiral.endpoint https://$ip_address:$port
 secure
 
 configureAdmiralStart ADMIRAL_DATA_LOCATION $data_dir
-configureAdmiralStart ADMIRAL_CERT_LOCATION $cert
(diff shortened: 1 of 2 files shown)


installer/packer/scripts/admiral/configure_admiral.sh
```diff
@@ -191,6 +189,22 @@ if [ ${harbor_deploy,,} == "true" ]; then
   echo "harbor.tab.url=https://${hostname}:${harbor_port}" > $data_dir/custom.conf
 fi
 
+# Copy files needed by Admiral into one directory
+config_dir=$data_dir/configs
+mkdir -p $config_dir
+admiral_psc_dir=/etc/vmware/psc/admiral
+
+cp $jks $config_dir
```
hmahmood: Are `$jks`, `$key`, `$cert` all non-empty this point always?

anchal-agrawal: Yes, the paths are set on lines 41-43.

Issue #384, opened 2017-06-22
Automate OVA upgrade process


frapposelli:
**User Statement:**

As a VIC customer, I'd like an automated procedure to upgrade my OVA deployment once a new version comes out, it should be as easy as downloading the new OVA and run a script or binary on my local machine.

**Details:**

The current upgrade process consists of some manual operations to be performed by the vSphere admin:

- Shutdown old VIC VM
- Deploy new OVA
- Detach old data disk
- Replace data disk in the newly deployed VIC VM with the old data disk
- Power on new VIC VM

These steps can be automated with `govc`, ideally the script/tool should:

- Fetch the existing OVF config from the running VIC VM
- Shut down the old VIC VM
- Deploy the new OVA, using the config fetched previously
- Before powering on the new OVA, copy the data disk from the old VIC VM over the empty data disk of the new VIC VM.
- Start the new VIC VM
- Notify the vSphere admin to remove the old VIC VM once the upgrade has been validated.

**Acceptance Criteria:**

The tool should automate all the vSphere-level operations at a very minimum.
Stretch goal is to automatically fetch and apply the OVF config.

andrewtchin:
Good points to consider for our docs https://docs.vmware.com/en/VMware-vSphere/6.5/com.vmware.vsphere.upgrade.doc/GUID-913F49FC-E0F4-4AEE-8848-A0FE34F05E1D.html

andrewtchin:
https://docs.vmware.com/en/VMware-vSphere/6.5/com.vmware.vsphere.upgrade.doc/GUID-34E5EB36-948A-42B5-949F-00412E1D1026.html

andrewtchin:
should have checklist to verify upgrade is successful

PR #466, opened 2017-07-27
Removing support for upgrade from 1.0 to 1.2.

stuclem:
Removing 1.0. to 1.2 as a supported upgrade path. I have left the upgrade process as it is, pending the completion of https://github.com/vmware/vic-product/issues/384.


stuclem:
Thanks @andrewtchin !

PR #475, opened 2017-07-31
remove kov

maplain:


skyscooby:
@andrewtchin @anchal-agrawal @maplain 
What happened to this project? This was going to bring the kubernetes API to VIC correct?

mdubya66:
Hi @skyscooby. The project has been replaced with the Pivotal Container Service (PKS) which was announced at VMworld. You can read about it here. https://www.vmware.com/radius/pivotal-container-services/

PR #492, opened 2017-08-01
Update Harbor startup

andrewtchin:


anchal-agrawal:
From @ywk253100 via Slack:

> and here should be another line to replace the data directory in docker-compose.clair.yml  https://github.com/vmware/vic-product/blob/master/installer/packer/scripts/provision_harbor.sh#L83

PR #537, opened 2017-08-08
Add params to Admiral's docker run command

anchal-agrawal:
This change adds two JAVA_OPTS values to the Admiral container,
per request from @lazarin.


anchal-agrawal:
Verified that Admiral and Harbor start with this change, and Harbor can be added to Admiral. Merging.

Issue #515, opened 2017-08-04
Revert all anchors to HTML tags


stuclem:
As pointed out by @anchal-agrawal in https://github.com/vmware/vic-product/pull/496, the Gitbook style of anchor (`{#anchor}`) does not render correctly in Github. So, we need to revert all anchors to HTML `a id` tags.

PR #559, opened 2017-08-11
Replaced Gitbook anchors with HTML anchors

stuclem:
Gitbook anchors (`{#anchor}`) work in Gitbook HTML but do no render correctly when viewing the Markdown files in Github. Reverted all anchors to HTML `<a id>` tags.

Fixes https://github.com/vmware/vic-product/issues/515.

@anchal-agrawal, as the reporter of the initial issue, you get the honour of reviewing the fix :)


stuclem:
Thanks @anchal-agrawal !

PR #561, opened 2017-08-11
Update photon packages and remove old kernel

andrewtchin:


installer/packer/scripts/filesystem_provisioning.sh
```diff
@@ -0,0 +1,37 @@
...
+# Create directories for PSC token
+mkdir -p /etc/vmware/psc/admiral
+mkdir -p /etc/vmware/psc/harbor
+mkdir -p /etc/vmware/psc/engine
+
+# Create Data Dir
+mkdir /data
+
+# Create filesystem on /dev/sdb to be mounted as /data
+parted -a optimal --script /dev/sdb 'mklabel gpt mkpart primary ext4 0% 100%'
```
anchal-agrawal: Is the `data.mount` service still needed after this change?

andrewtchin: yes this only moves half of what was in package provisioning to a new shell provisioner due to reboot required after updating the kernel

Issue #316, opened 2017-06-15
Document docker commit


stuclem:
Per https://github.com/vmware/vic/issues/5253 and https://github.com/vmware/vic/issues/4743

stuclem:
Confirmed with @sflxn that this is definitely in 1.2.

PR #576, opened 2017-08-15
Documenting support for docker commit

stuclem:
Looks like we only mention `docker commit` in the table of supported Docker commands. Mentioned that you can only run it on stopped containers.

Fixes https://github.com/vmware/vic-product/issues/316.

@sflxn, can you please review? Are there any other caveats that we need to mention? Thanks!


stuclem:
Thanks @sflxn !

Issue #557, opened 2017-08-11
Document updated behaviour of vic-machine inspect config


stuclem:
Per https://github.com/vmware/vic/issues/5985, implemented in PR https://github.com/vmware/vic/pull/5993. At present, we document that config must be after the options for inspect, which will no longer be the case after this PR is merged.

stuclem:
https://github.com/vmware/vic/pull/5993 is merged

stuclem:
In fact, if you specify config after the other options, the operation now fails:

```
$ vic-machine-windows inspect --target 'Administrator@vsphere.local:password'@10.153.199.2 --thumbprint FD:38:47:57:88:D9:AC:BE:69:1A:E9:2F:FA:C4:62:BD:82:69:48:FB --id vm-250 config
Aug 15 2017 12:59:22.902+02:00 INFO  ### Inspecting VCH ####
Aug 15 2017 12:59:24.503+02:00 ERROR Failed to verify certificate for target=10.153.199.2 (thumbprint=FD:38:47:57:88:D9:AC:BE:69:1A:E9:2F:FA:C4:62:BD:82:69:48:FB)
Aug 15 2017 12:59:24.503+02:00 ERROR Inspect cannot continue - failed to create validator: x509: certificate signed by unknown authority
Aug 15 2017 12:59:24.503+02:00 ERROR --------------------
Aug 15 2017 12:59:24.503+02:00 ERROR vic-machine-windows.exe inspect config failed: inspect failed
```

PR #578, opened 2017-08-15
Updated inspect config to reflect new behaviour.

stuclem:
Fixes https://github.com/vmware/vic-product/issues/557.

@emlin is this OK? Thanks!


stuclem:
Actually, this is a very straightforward fix that I want to merge so that I can generate HTML. @alextopuzov can you please take a look?

PR #582, opened 2017-08-15
Implement support for creating default users in the OVA

lcastellano:


installer/fileserver/tasks.go
```diff
@@ -65,6 +65,12 @@ func registerWithPSC(ctx context.Context) error {
 	}
 	admiralPort := ovf.Properties["management_portal.port"]
 
+	// Out of the box users
+	defPrefix, foundPrefix := ovf.Properties["default_users.def_user_prefix"]
+	defPassword, foundPassword := ovf.Properties["default_users.def_user_password"]
+
+	log.Infof("PSC Out of the box users. Prefix: %s Password %s", defPrefix, defPassword)
```
andrewtchin: I don't think we're supposed to log passwords

installer/packer/vic-unified.ovf
```diff
@@ -286,6 +286,18 @@ EVALUATION LICENSE. If You are licensing the Software for evaluation purposes, Y
         <Description>Specifies the port on which fileserver will be published.</Description>
       </Property>
     </ProductSection>
+    <ProductSection ovf:class="default_users" ovf:required="false">
```
andrewtchin: Do we always want to create OOTB users or should there be an enable/disable?

andrewtchin:
does this close https://github.com/vmware/vic/issues/4729?

installer/fileserver/tasks.go
```diff
@@ -83,6 +89,19 @@ func registerWithPSC(ctx context.Context) error {
...
+			if foundPrefix {
+				arg := "--defaultUserPrefix=" + defPrefix;
+				cmdArgs = append(cmdArgs, arg)
+			}
+
+			if foundPassword {
+				arg := "--defaultUserPassword=" + "\\'" + defPassword + "\\'";
+				cmdArgs = append(cmdArgs, arg)
+			}
+			log.Infof("PSC register arguments: %v", cmdArgs)
```
anchal-agrawal: We shouldn't log the args since they contain vSphere credentials.

installer/packer/vic-unified.ovf
lcastellano: If you remove either the prefix or the password they shouldn't be created. I have not tested that case I will try now.

installer/fileserver/tasks.go
lcastellano: Yes, I agree. That log statement was for debugging it should be removed.

installer/packer/vic-unified.ovf
lcastellano: Sorry I was mistaken, in the current implementation you must have a prefix and a password, therefore there is no way not to have default users. I think we can leave it like this for 1.2. You can always chose not to use them. The users are created in VSphere but without any VSphere roles.

andrewtchin: You can add a bool like this 
```
      <Property ovf:key="gc_enabled" ovf:type="boolean" ovf:userConfigurable="true" ovf:value="false">
        <Label>3.3. Garbage Collection</Label>
        <Description>When setting this to true, registry performs garbage collection everytime it boots up.</Description>
      </Property>
``` 
Then in the script test like this:
```
gc_enabled=$(ovfenv --key registry.gc_enabled)
if [ "${gc_enabled,,}" == "true" ]; then
```

installer/fileserver/tasks.go
lcastellano: Oh yes, good catch, I forgot to remove the logging statements

installer/packer/vic-unified.ovf
```diff
@@ -278,6 +278,22 @@ EVALUATION LICENSE. If You are licensing the Software for evaluation purposes, Y
...
```
lcastellano: Code changes overview:
**vic-unified.ovf** contains the new template customization params for out-of-the-box users

Section 7, contains 3 fields: one boolean to enable the out-of-the-box user creation, one string to use as the prefix and one password to use as the common password for all the out-of-the-box users; there are defaults for the prefix and the password. 

screen shot 2017-08-22 at 10 36 23 am

Out of the box user creation is skipped in the following cases:
1) create_def_users is set to False,
2) def_user_prefix is empty
3) def_user_password is empty,

Otherwise the out-of-the-box users are created.

The value of those ovfenv variables are checked in **tasks.go** and in **add_default_users.sh**. 
The creation consists of two steps:
1) adding the users to vSphere (PSC)
2) providing the users with roles in Admiral
The first step is achieved by providing additional options to the the psc-registration command. The command is run from Go, the code changes are in **tasks.go**. 
The second step is achieved by **add_default_users.sh**. The script requires three variables to be initialized, this is achieved by having the script **configure_admiral.sh** modifying its content (similar to **start_admiral.sh**). The script configure_admiral.sh has changed to include the modification of **add_default_users.sh** along with **start_admiral.sh**.
Finally a new systemd service file has been added (**admiral_default_users.service**) along with triggering its activation in **admiral_startup.service**.

The script **add_default_users.sh** first checks variables in ovfenv. If the variables require to skip out-of-the-box user creation, addition to Admiral is skipped.
The script **add_default_users.sh** must wait for Admiral and the Admiral Token to be available. For that reason it has two loops where it first check for the existence of the token and then makes sure Admiral is ready by pinging it. When all is ready the scripts calls two Admiral REST API to provide the users with their correct role.

Here is the log output of the execution of the script:
Aug 21 21:23:52 localhost.localdomain bash[1000]: add_default_users: True, vicdef
Aug 21 21:23:52 localhost.localdomain bash[1000]: add_default_users wating for token
Aug 21 21:23:52 localhost.localdomain bash[1000]: add_default_users loaded token
Aug 21 21:23:52 localhost.localdomain bash[1000]: add_default_users ping result: 404
Aug 21 21:23:52 localhost.localdomain bash[1000]: add_default_users ping failed
Aug 21 21:23:53 localhost.localdomain bash[1000]: add_default_users ping result: 404
Aug 21 21:23:53 localhost.localdomain bash[1000]: add_default_users ping failed
Aug 21 21:23:55 localhost.localdomain bash[1000]: add_default_users ping result: 404
Aug 21 21:23:55 localhost.localdomain bash[1000]: add_default_users ping failed
Aug 21 21:23:56 localhost.localdomain bash[1000]: add_default_users ping result: 404
Aug 21 21:23:56 localhost.localdomain bash[1000]: add_default_users ping failed
Aug 21 21:23:57 localhost.localdomain bash[1000]: add_default_users ping result: 200
Aug 21 21:23:57 localhost.localdomain bash[1000]: add_default_users successful ping
Aug 21 21:23:57 localhost.localdomain bash[1000]: add_default_users added cloud-admin
Aug 21 21:23:57 localhost.localdomain bash[1000]: add_default_users added project-admin

The script **add_default_users.sh** can be executed multiple times.

installer/fileserver/tasks.go
```diff
@@ -83,6 +91,18 @@ func registerWithPSC(ctx context.Context) error {
 			"--configDir=" + pscConfDir,
 		}
 
+		if client == "admiral" && foundCreateUsers == true && strings.ToLower(defCreateUsers) == "true" {
```
anchal-agrawal: `foundCreateUsers == true` -> `foundCreateUsers`

installer/fileserver/tasks.go
```diff
@@ -83,6 +91,18 @@ func registerWithPSC(ctx context.Context) error {
...
+			if foundPrefix && defPrefix != "" {
+				arg := "--defaultUserPrefix=" + defPrefix;
```
anchal-agrawal: Trailing semicolon not necessary.

installer/packer/scripts/admiral/add_default_users.sh
```diff
@@ -0,0 +1,151 @@
...
+
+echo "add_default_users wating for token"
+token_tries=0
+while true ; do
+    if [ -f $token_file ]; then
+        break;
+    fi
+    ((token_tries++))
+    sleep 1
+    if [ ${current_tries} -eq  60 ]; then
```
anchal-agrawal: Should `token_tries` be used instead of `current_tries ` here?

installer/packer/scripts/systemd/admiral/admiral_default_users.service
```diff
@@ -0,0 +1,12 @@
+[Unit]
+Description=Admiral Startup Configuration
```
anchal-agrawal: Please change the description to say something like `Admiral Default Users`.

installer/packer/scripts/admiral/add_default_users.sh
```diff
@@ -0,0 +1,151 @@
...
+echo "add_default_users loaded token"
+
+tenant=`get_property $psc_prop_file "tenant"`
+defuser_prefix=`get_property $psc_prop_file "default-user-prefix"`
+admiral_url=`get_property $psc_prop_file admiral-url`
+# remove backslashes
+admiral_url=`echo $admiral_url | sed 's/\\\//g'`
+
+cloud_admin_name=$defuser_prefix
+cloud_admin_name+="-cloud-admin"
```
hmahmood: Can `$defuser_prefix` be empty? If yes, then the user name will begin with `-`.

lcastellano: The prefix is saved into the config file by the register command. The register command takes it from the ovfenv variable: default_users.def_user_prefix. At line 35 of add_default_users.sh the value of default_users.def_user_prefix is checked against being empty.

installer/fileserver/tasks.go
```diff
@@ -83,6 +91,18 @@ func registerWithPSC(ctx context.Context) error {
...
 		}
 
+		if client == "admiral" && foundCreateUsers && strings.ToLower(defCreateUsers) == "true" {
+			if foundPrefix && defPrefix != "" {
+				arg := "--defaultUserPrefix=" + defPrefix;
+				cmdArgs = append(cmdArgs, arg)
+			}
+
+			if foundPassword && defPrefix != "" && defPassword != "" {
+				arg := "--defaultUserPassword=" + defPassword
```
anchal-agrawal: Edge-case: is it possible that the PSC command runs with either `--defaultUserPrefix=foo` or `--defaultUserPrefix=bar`, but not both? How does the PSC code handle that case, if it is possible?

installer/fileserver/tasks.go
```diff
@@ -83,6 +91,18 @@ func registerWithPSC(ctx context.Context) error {
 			"--configDir=" + pscConfDir,
 		}
 
+		if client == "admiral" && foundCreateUsers && strings.ToLower(defCreateUsers) == "true" {
+			if foundPrefix && defPrefix != "" {
+				arg := "--defaultUserPrefix=" + defPrefix;
```
anchal-agrawal: Looks like the trailing semicolon is still present.

installer/fileserver/tasks.go
anchal-agrawal: Clarified via Slack that the PSC command handles this.

andrewtchin:
@lcastellano see this issue and see if we can address it in the PR https://github.com/vmware/vic-product/issues/609

Issue #198, opened 2017-05-02
Document that you cannot delete a container VM from VC


stuclem:
Per https://github.com/vmware/vic/issues/2928, which is implemented in https://github.com/vmware/vic/pull/4965.

stuclem:
See also https://github.com/vmware/vic/issues/5238

PR #597, opened 2017-08-18
Mentioned that you cannot delete container VMs in client.

stuclem:
Fixes https://github.com/vmware/vic-product/issues/198.

@dougm can you please review?


stuclem:
Thanks @dougm !

PR #598, opened 2017-08-18
Removed OVA option to provide custom certs for registry.

stuclem:
Noticed in vic-42ddaaa6.ova that the options to add SSL certs for VIC registry have been removed from the OVA. 

No associated issue. @anchal-agrawal are you a good candidate to review this in @andrewtchin 's absence? Thanks!


stuclem:
Thanks @anchal-agrawal !

Issue #319, opened 2017-06-15
Document environment variables for vic-machine


stuclem:
Per https://github.com/vmware/vic/issues/5246, which is implemented in https://github.com/vmware/vic/pull/5269

stuclem:
See also https://github.com/vmware/vic/pull/5099.

PR #599, opened 2017-08-18
Added doc about vic-machine environment variables.

stuclem:
Fixes https://github.com/vmware/vic-product/issues/319.

@morris-jason can you please review? Thanks!


stuclem:
Thanks @morris-jason !

PR #607, opened 2017-08-18
Fix getting started layout issues

### LINKED ISSUE #555: OVA  - Disable login button once you click on it
With my setup running on onecloud, it can be slow at times. When I put in my creds and click the "Login" button, it takes a few seconds at which point I keep clicking the button multiple times. 

When the login button is clicked once, it should be disabled until response to that click is received.

### LINKED ISSUE #567: Add some progress/activity indication after first login to appliance
When you first go to https://**vic-appliance-address**:9443 and enter the vCenter Server credentials, the only indication that anything is happening is the Connecting... message in the browser tab.

login

It would be nice if there were some indication in the login screen to indicate that the login/configuration is in progress.

### LINKED ISSUE #591: Getting Started page: Fix UX of the 'initialize VIC App..' button
Currently a Clarity outline button is displayed to the right of the Getting Started page title if the continued installation process from this page fails. (see image)

screen shot 2017-08-17 at 11 29 11 am

(Note: The Document button should be aligned to the right with a gutter of 24px from the right edge of the view-port and 24px under the header or alert component)

The correct interaction should be to show a Clarity alert, inform the user of the issue and offer the action to fix this issue in the alert. (see image)

screen shot 2017-08-17 at 3 12 55 pm

As a precaution, the Clarity flat button used in this alert should also appear at the bottom of the Getting Started page, in case there is any need (say a request from Support) for the vSphere Admin to restart the initialization again when the alert is not shown. (see image above)

Ref: Specific InVision mockups at https://vmware.invisionapp.com/share/QYD37KV84

cc: @andrewtchin @stuclem

cristianfalcone:
Fixes #555 
Fixes #567 
Fixes #591

### DIFF
--- a/installer/fileserver/html/css/style.css
+++ b/installer/fileserver/html/css/style.css
@@ -1,8 +1,4 @@
-/* Header / navigation section */
-
-.main-nav .header {
-    background-color: #0B6C8F;
-}
+/* Header/Navigation section */
 
 .header .github {
     position: absolute;
@@ -24,11 +20,10 @@
     font-size: 20px
 }
 
-
 /* Clarity UI overrides */
 
 .content-area {
-    padding: 0px !important;
+    padding: 0 !important;
 }
 
 .card.transparent {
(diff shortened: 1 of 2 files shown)


cristianfalcone:
Tested in latest Google Chrome, latest Firefox and IE11.

PR #621, opened 2017-08-23
Added info on log in to MP for 1st time

alextopuzov:
Part of clarifying the users story in 1.2.
Added a concept topic in the MP doc, added to Summary.

### DIFF
--- a/docs/user_doc/vic_cloud_admin/logging_in_MP.md
+++ b/docs/user_doc/vic_cloud_admin/logging_in_MP.md
@@ -7,7 +7,7 @@ If you don't know the port number, you can access the portal by going to http://
 ## Default User Access to the Management Portal ##
 
 The role that has full permissions for vSphere Integrated Containers is the cloud administrator role. 
-By default, the cloud administrator role is assigned to the Administrators group for vCenter Server during the installation of vSphere Integrated Containers. Every user that is a member of that group in the Platform Services Controller can access the Management Portal as cloud administrator. After you log in as a cloud administrator, you can give other users access to vSphere Integrated Containers by assigning them roles to projects.
+By default, the cloud administrator role is assigned to the Administrators group for vCenter Server during the installation of vSphere Integrated Containers. Every user that is a member of that group in the Platform Services Controller can access the Management Portal as cloud administrator. After you log in as a cloud administrator, you can give other users access to vSphere Integrated Containers by assigning them roles in projects.
 
 Optionally, you can log in with a user that you created during the OVA deployment, if you used that option.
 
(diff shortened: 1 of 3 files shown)


docs/user_doc/vic_cloud_admin/logging_in_MP.md
```diff
@@ -0,0 +1,14 @@
+# Logging In to the Management Portal #
+
+You can access the Management Portal in a web browser by entering the vSphere Integrated Containers instance IP address and the port that you specified for the portal during the deployment. By default the port number is *8282*.
```
stuclem: "vSphere Integrated Containers instance IP address" --> "vSphere Integrated Containers appliance IP address"

docs/user_doc/vic_cloud_admin/logging_in_MP.md
```diff
@@ -0,0 +1,14 @@
...
+
+If you don't know the port number, you can access the portal by going to http://<i>vic_appliance_address</i> and following the **Go to the vSphere Integrated Containers Management Portal** link.
+
+## Default User Access to the Management Portal ##
+
+The role that has full permissions for vSphere Integrated Containers is the cloud administrator role. 
+By default, the cloud administrator role is assigned to the Administrators group for vCenter Server during the installation of vSphere Integrated Containers. Every user that is a member of that group in the Platform Services Controller can access the Management Portal as cloud administrator. After you log in as a cloud administrator, you can give other users access to vSphere Integrated Containers by assigning them roles to projects.
```
stuclem: "by assigning them roles to projects" --> "by assigning them roles in projects"

docs/user_doc/vic_cloud_admin/logging_in_MP.md
```diff
@@ -0,0 +1,14 @@
...
+
+Optionally, you can log in with a user that you created during the OVA deployment, if you used that option.
```
stuclem: We need to confirm that this does indeed end up being the case.

docs/user_doc/vic_cloud_admin/logging_in_MP.md
alextopuzov: fixed

docs/user_doc/vic_cloud_admin/logging_in_MP.md
alextopuzov: fixed

docs/user_doc/vic_cloud_admin/logging_in_MP.md
```diff
@@ -0,0 +1,14 @@
+# Logging In to the Management Portal #
+
+You can access the Management Portal in a web browser by entering the vSphere Integrated Containers appliance IP address and the port that you specified for the portal during the deployment. By default the port number is *8282*.
```
stuclem: The users who log into the MP are not necessarily going to be the same as the vSphere admin who deployed the OVA, so "the port that you specified" is probably not  going to be "you".  I think that we can bypass the whole problem of the OVA using non-default ports by just sending people to the Getting Started page and telling them to click the link. This way, we don't need to mention port numbers at all.

stuclem:
I confirmed in today's build that the option to deploy the appliance with OOTB users is indeed implemented, so this is now good to go.

PR #622, opened 2017-08-23
updating product to 1.2 license

mdubya66:


NOTICE
```diff
@@ -0,0 +1,12 @@
+NOTICE
+
+vSphere Integrated Containers
+
+Copyright � 2017 VMware, Inc. All Rights Reserved.
```
anchal-agrawal: The copyright character looks off.

cgtexmex: (c) rendering as a _(?)_ -- feature or bug?

mdubya66: no idea. as it was delivered from legal

mhagen-vmware: that is just in github, if you click view on the file it renders correctly

Issue #252, opened 2017-05-26
Unified OVA still shows the old version number in the OVF deploy template summary before clicking finish to deploy


dbarkelew:
VIC 1.1.1 OVA shows 1.1.0 version in the details confirmation screens before clicking finish to start deploying the unified OVA. 

Seems strictly a cosmetic issue with the OVA packing. Versions after deployment match correctly for the 1.1.1 release as expected.

Bug 1881541

andrewtchin:
https://github.com/vmware/vic/issues/5192

PR #625, opened 2017-08-23
OVA build sets version programmatically

jonathanibers:
Building an ova file now sets the version number by replacing
"--version--" with the version set by the BUILD_VICENGINE_REVISION
environment variable.

Fixes: #252

### DIFF
--- a/installer/packer/vic-unified.ovf
+++ b/installer/packer/vic-unified.ovf
@@ -170,13 +170,16 @@ EVALUATION LICENSE. If You are licensing the Software for evaluation purposes, Y
             Version is the actual product version in the
             form X.X.X.X where X is an unsigned 16-bit integer.
 
+            This is set by the $BUILD_VICENGINE_REVISION
+            environment variable.
+
             FullVersion is a descriptive version string
             including, for example, alpha or beta designations
             and other release criteria.
         -->
 
 
-      <Version>1.1.1.1</Version>
+      <Version>--version--</Version>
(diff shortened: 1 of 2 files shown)


installer/packer/vic-unified.ovf
```diff
@@ -176,7 +176,7 @@ EVALUATION LICENSE. If You are licensing the Software for evaluation purposes, Y
         -->
```
anchal-agrawal: We should add to this comment that the version is set by the `$BUILD_VICENGINE_REVISION` env variable.

jonathanibers: Updated

andrewtchin:
@jonathanibers could you show an example of what the resulting version string looks like?

andrewtchin:
@jonathanibers Also is it possible for the version in the ovf file to contain letters? Like if we wanted to include the commit hash in the version?

jonathanibers:
screenshot from 2017-08-23 14-41-40

Text of ova file:

`
      <Version>1.1.1</Version>`

jonathanibers:
@andrewtchin That should be possible (in theory), as long as there's a way for the Makefile to access the commit hash.

andrewtchin:
So does that mean we should set FullVersion with the string that has the commit hash?

jonathanibers:
The FullVersion currently isn't set with anything (as far as I can tell) but I can look into setting it to the commit hash.

Issue #626, opened 2017-08-23
OVA Console version


andrewtchin:
![screen shot 2017-08-23 at 3 13 48 pm](https://user-images.githubusercontent.com/1649165/29636558-a9e90846-8816-11e7-9599-a1fad93a12e0.jpg)
Since we're not using Drone build, fix OVA console version to be `<tag>-<short commit hash>`

jonathanibers:
![ova console](https://user-images.githubusercontent.com/30810490/29639724-5db2fbf2-8821-11e7-9444-644af11ed5b9.png)

New version

PR #652, opened 2017-08-25
Removes the build number from OVA console version

jonathanibers:
The OVA console version now diplays <tag>-<commit-hash>
on the top line.

Fixes: #626


jonathanibers:
ova-console-new

Issue #383, opened 2017-06-22
Document that OVA checks network config before deployment


stuclem:
Per https://github.com/vmware/vic/issues/4987 and PR https://github.com/vmware/vic-product/pull/350.

PR #665, opened 2017-08-28
Documenting network status console for appliance VM

stuclem:
Fixes https://github.com/vmware/vic-product/issues/383

@morris-jason is this enough?


stuclem:
Thanks @morris-jason !

PR #671, opened 2017-08-28
Use abs path for timestamp file on OVA

anchal-agrawal:
Towards https://github.com/vmware/vic/issues/4665


anchal-agrawal:
Closing for now - I'll revisit this.

Issue #654, opened 2017-08-28
Document that default debug level is 1, not 0


stuclem:
If we do merge https://github.com/vmware/vic/pull/6102 in 1.2, we should update the description of `vic-machine create debug` accordingly. Including in the 1.2 release until confirmation of whether this is in or out.

sgairo:
Looks like its in :)

stuclem:
Yup, thanks @sgairo. Moving to To Do.

PR #678, opened 2017-08-29
Updated default debug level from 0 to 1.

stuclem:
Fixes https://github.com/vmware/vic-product/issues/654

@emlin can you please review?


docs/user_doc/vic_vsphere_admin/configure_vch.md
```diff
@@ -253,7 +253,7 @@ This example increases the level of debugging to level 3, either on a VCH that i
     --id <i>vch_id</i>
     --debug 3</pre>
 
-This example sets the `--debug` option to 0, to disable debug mode on a VCH that is already running at any level of debug mode.
+This example sets the `--debug` option to 0, to disable debug mode on a VCH that is already running at any level of debug mode. By default, VCHs run with debugging level 1.
```
emlin: for vic-machine configure command, no default value for --debug option. If no --debug supplied, the old debug option will not be changed.

stuclem: Thanks @emlin, I was trying to show that VCHs are created with debug 1 by default, but I agree that this is not the right place to say that and it wasn't stated very clearly. Hopefully this is clarified in

Issue #613, opened 2017-08-22
Install plugin doc: Minor enhancements


ali5ter:
Some thoughts to enhance [documentation titled. "Install the Client Plug-Ins on a vCenter Server Appliance"](https://vmware.github.io/vic-product/assets/files/html/1.2/vic_vsphere_admin/plugins_vcsa.html):
* Pre-reqs bullet that reads "Go to the vCenter Server Appliance Management Interface (VAMI) at https://vcsa_address:5480, click Access, and make sure that Bash Shell is enabled." might usefully state that the console credentials are used to access the VAMI. Not sure we can assume vSphere Admin should already know this. 
* Procedure 1 could usefully add that to access the bash shell, after you login you type 'shell' to start bash before you can enter the curl command. In other words you don't directly start a shell when you log in. Example of this experience can be seen 9m20s into [this screencast](https://onevmw-my.sharepoint.com/personal/bowena_vmware_com/_layouts/15/guestaccess.aspx?guestaccesstoken=EohssxOtCXpoNrRRKbsjwlq3U9N8WTwdgNZgkosoAdE%3d&docid=2_19aaa26f045084b7f9f5016d79cfbffa4&rev=1).
* Procedure 2 explains the curl command. Would be useful to explain that the reader can find out the name of the vic binary by navigating to the file-server url, https://appliance_ip:9443/files/
* Procedure 3 explains how to restart the vsphere-ui and vsphere-client services. Since no --restart option is supplied, the two commands could be consolidated into one command line that the vSphere Admin can cut and paste, i.e. service-control --stop vsphere-ui && service-control --start vsphere-ui and service-control --stop vsphere-client && service-control --start vsphere-client

stuclem:
Thanks @ali5ter, these are useful observations.

pdaigle:
I would also recommend documenting the requirement for `awk` for both the `install.sh` and `upgrade.sh` scripts. 

For example, `awk` is not installed by default on Photon OS.

Both install.sh and upgade.sh fail if `awk` is not present:

![image](https://user-images.githubusercontent.com/8586715/29618564-92b41e84-87e6-11e7-940b-525ed49a8887.png)

stuclem:
@pdaigle we don't support Photon OS as a platform for running `vic-machine`, so by extension I don't think that we would support it for running the plug-in scripts:

https://vmware.github.io/vic-product/assets/files/html/1.2/vic_vsphere_admin/vic_installation_prereqs.html#platforms

[...]

Issue #612, opened 2017-08-22
Install plugin doc: curl command needs -L option


ali5ter:
[Documentation titled. "Install the Client Plug-Ins on a vCenter Server Appliance"](https://vmware.github.io/vic-product/assets/files/html/1.2/vic_vsphere_admin/plugins_vcsa.html), procedure 2 explains the curl command to download the vic binary from the appliance (OVA) file-server

The '-L' option is missing. This makes sure curl negotiates the redirect that was introduced in v1.2 to get to the Appliance file-server.

e.g. curl -kL http://10.17.125.48/files/vic_13202.tar.gz -o vic_13202.tar.gz

PR #698, opened 2017-08-30
Comments from @ali5ter about installing the plugins.

stuclem:
- Fixes https://github.com/vmware/vic-product/issues/613
- Fixes https://github.com/vmware/vic-product/issues/612

@ali5ter, since you reported the above issues, can you please review? @jooskim too, obviously :)


stuclem:
Thanks @ali5ter and @jooskim !

PR #701, opened 2017-08-30
Added awk as a prereq for plugin and merged ui commands

stuclem:
Fixes https://github.com/vmware/vic-product/issues/613.

@jooskim is this OK?


stuclem:
Thanks @jooskim !

PR #709, opened 2017-08-30
Add DCH to OVA

andrewtchin:
Fixes https://github.com/vmware/vic/issues/5463
screen shot 2017-08-30 at 6 15 08 pm

### DIFF
--- a/installer/packer/scripts/system_settings.sh
+++ b/installer/packer/scripts/system_settings.sh
@@ -40,7 +40,7 @@ systemctl enable fileserver_startup.service fileserver.service
 systemctl enable engine_installer_startup.service engine_installer.service
 
 # Clean up temporary directories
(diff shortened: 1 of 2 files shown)


mdubya66:
Forgive my ignorance but does this happen during OVA build time or during OVA install?

andrewtchin:
during build time

installer/packer/scripts/system_settings.sh
```diff
@@ -41,6 +41,7 @@ systemctl enable engine_installer_startup.service engine_installer.service
 
 # Clean up temporary directories
 rm -rf /var/tmp/harbor
```
anchal-agrawal: This line can be nuked.

andrewtchin: lol thanks fixed

PR #727, opened 2017-09-04
More upgrade doc updates

stuclem:
Fixes https://github.com/vmware/vic-product/issues/733 and removes the arguments from running the script.

@andrewtchin can you please review? Thanks!


stuclem:
I have added the extra step to the upgrade procedure, as provided by @lazarin in https://github.com/vmware/admiral/issues/191. However, in my opinion, while I have tested it and it works, it does not look great as the officially-approved upgrade method. If there is any way that we can get the fix into 1.2 rather than obliging customers to perform this step, I think it would be a lot cleaner.

stuclem:
Thanks @andrewtchin. Merging this one, but I do want to check the upgrade again once https://github.com/vmware/harbor/issues/3163 is fixed, so that I can confirm what data is migrated and what post-upgrade steps are required to make your new 1.2 instance identical to the 1.1.1 setup.

PR #739, opened 2017-09-06
Updating systemctl commands

stuclem:
No associated issue, but I noticed in https://github.com/vmware/admiral/issues/191 that that the systemctl commands used by @lazarin are different to what I have documented.

@andrewtchin is this now correct? Should there be another command for the Demo VCH service?


andrewtchin:
They both work. The Demo VCH Installer Wizard is `engine_installer.service` if you wanted to add that

stuclem:
Thanks @andrewtchin !

Issue #324, opened 2017-06-15
Document dch-photon


stuclem:
Per https://github.com/vmware/vic/issues/4852. This might become an Epic.

stuclem:
See also https://github.com/vmware/vic/issues/4868

stuclem:
Also https://github.com/vmware/vic/issues/4859

stuclem:
Per https://github.com/vmware/vic-product/pull/544, DinV is now called `dch-photon`.

stuclem:
Comment from @andrewtchin via Slack:

Andrew Chin [1:40 AM] 
i was testing dch-photon and wanted to make sure we have documented that the vch is required to be deployed with --registry-ca to be able to use images from harbor
[1:41] 
also depending on this result we may need a doc https://github.com/vmware/admiral/issues/192

andrewtchin:
I confirmed with mavery that ANY container that requires anonymous volumes requires a VCH that has a volume store with the label `default` specified otherwise the error from vmware/admiral#192 happens. This should be documented in the DCH docs and with other volume docs

stuclem:
Ok, thanks @andrewtchin

PR #773, opened 2017-09-08
Documenting dch-photon

stuclem:
Fixes https://github.com/vmware/vic-product/issues/324.

@frapposelli is on holiday. @mhagen-vmware is anyone from your team a good candidate to review this? 

Thanks!


stuclem:
@corrieb can you please take a look?

stuclem:
Incidentally, I'll be adding a second procedure on how to use DinV as a fully fledged Docker host

stuclem:
I have written up steps that I am convinced _should_ work. @corrieb has tried them and says that they do. Maybe dch-photon doesn't like me running Docker on Windows.

@corrieb can you please review now? @mhagen-vmware can you have someone from QE go through the procedure?

Thanks!

stuclem:
Incidentally @rajanashok , one thing that I am not sure about is whether to deploy the `dch-photon` container with the `-tls` option or `-tlsverify` (these are DCH deployment options, not VCH options - I deployed my VCH with `--no-tlsverify`). One thing that's for sure is that you have to have one of them. @corrieb says that these options should have no effect on the connection from the DCH container to the Registry, but if I omit them (or use the DCH option `-insecure-registry`) the login attempt to the Registry from DCH fails with `error during connect: Post https://10.153.199.225:12376/v1.30/auth: dial tcp 10.153.199.225:12376: connectex: No connection could be made because the target machine actively refused it.` where 10.153.199.225 is the VCH.

If I specify either one of `-tls` or `-tlsverify` I get `Error response from daemon: Get https://10.153.199.229/v1/users/: net/http: TLS handshake timeout` when I attempt to log in to the registry.

stuclem:
Here's a cleaner version of the file: https://github.com/stuclem/vic-product/blob/dchphoton/docs/user_doc/vic_app_dev/build_push_images.md

Issue #771, opened 2017-09-08
PSC register failure: com.vmware.vim.vmomi.client.common.UnexpectedStatusCodeException: Unexpected status code: 400


lazarin:
This is a copy of an issue [196](https://github.com/vmware/admiral/issues/196) initially opened in Admiral product:

----

Found by a customer in OVA RC3 build https://storage.googleapis.com/vic-product-ova-builds/vic-v1.2.0-rc3-ce9fca67.ova

Upon running `admiral-auth-psc-1.2.0-SNAPSHOT-command.jar --command=register --tenant=<non " .local" domain>  ... (other args)`, it fails with:

```
[0][I][2017-09-07T19:53:20.378Z][1][SsoPscCommand][register][Initializing...]
[1][I][2017-09-07T19:53:20.385Z][1][SsoPscCommand][register][Registering...]
[2][I][2017-09-07T19:53:20.409Z][1][SsoPscCommand][deleteFile][Removal of file: '/etc/vmware/psc/harbor/psc-config.keystore' result: true]
[3][I][2017-09-07T19:53:31.070Z][1][VmodlContextImpl$NonValidatingClassPathXmlApplicationContext][prepareRefresh][Refreshing com.vmware.vim.vmomi.core.types.impl.VmodlContextImpl$NonValidatingClassPathXmlApplicationContext@59e5ddf: startup date [Thu Sep 07 19:53:31 UTC 2017]; root of context hierarchy]
[4][I][2017-09-07T19:53:31.097Z][1][XmlBeanDefinitionReader][loadBeanDefinitions][Loading XML bean definitions from class path resource [com/vmware/vim/binding/vmodl/context_v2.xml]]
[5][I][2017-09-07T19:53:31.294Z][1][VmodlContextImpl$NonValidatingClassPathXmlApplicationContext][doClose][Closing com.vmware.vim.vmomi.core.types.impl.VmodlContextImpl$NonValidatingClassPathXmlApplicationContext@59e5ddf: startup date [Thu Sep 07 19:53:31 UTC 2017]; root of context hierarchy]
[6][I][2017-09-07T19:53:31.296Z][1][VmodlContextImpl][loadVmodlPackage][Package com.vmware.vim.binding.vmodl loaded in 243 millis]
[...]

PR #774, opened 2017-09-08
Add support for external PSC in Getting Started page

anchal-agrawal:
This change allows the user to specify an external PSC instance
(not hosted on the vCenter host) as an optional input field in the
Getting Started page's login modal.

login_modal

Towards #771


andrewtchin:
@lcastellano confirmed the need for a separate domain field and will confirm with PSC team if we need separate user/pass fields

andrewtchin:
We will add additional fields as needed in a later issue. Getting this merged so we can get a dev build out

PR #779, opened 2017-09-10
Add BridgeNetworkRange to Demo VCH Installer Wizard

amalashin:
My main infrastructure network is 172.16.100.0 which conflicting with bridge network inside VCH. Thus deployment with Demo VCH Install Wizard is not working (all traffic for external purposes goes throughout bridge network instead of client network). This request is to add BridgeNetworkRange to the Demo Installer Wizard.

### DIFF
--- a/installer/engine_installer/engine_installer.go
+++ b/installer/engine_installer/engine_installer.go
@@ -44,6 +44,7 @@ type EngineInstallerConfigOptions struct {
 type EngineInstaller struct {
 	loginInfo       lib.LoginInfo
 	BridgeNetwork   string `json:"bridge-net"`
+	BridgeNetworkRange   string `json:"bridge-net-range"`
 	PublicNetwork   string `json:"public-net"`
 	ImageStore      string `json:"img-store"`
 	ComputeResource string `json:"compute"`
@@ -60,6 +61,7 @@ type AuthHTML struct {
 // ExecHTMLOptions contains fields for html templating in exec.html
 type ExecHTMLOptions struct {
 	BridgeNetwork   template.HTML
+	BridgeNetworkRange   template.HTML
 	PublicNetwork   template.HTML
 	ImageStore      template.HTML
 	ComputeResource template.HTML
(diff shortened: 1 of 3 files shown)


installer/engine_installer/engine_installer.go
```diff
@@ -44,6 +44,7 @@ type EngineInstallerConfigOptions struct {
 type EngineInstaller struct {
 	loginInfo       lib.LoginInfo
 	BridgeNetwork   string `json:"bridge-net"`
+	BridgeNetworkRange   string `json:"bridge-net-range"`
```
andrewtchin: Looks like `gofmt` might need to be run for formatting in this file

installer/engine_installer/engine_installer.go
```diff
@@ -147,6 +149,7 @@ func (ei *EngineInstaller) buildCreateCommand(binaryPath string) {
 	createCommand = append(createCommand, []string{"--name", ei.Name}...)
 	createCommand = append(createCommand, []string{"--public-network", ei.PublicNetwork}...)
 	createCommand = append(createCommand, []string{"--bridge-network", ei.BridgeNetwork}...)
+	createCommand = append(createCommand, []string{"--bridge-network-range", ei.BridgeNetworkRange}...)
```
andrewtchin: Can you put an `if` to exclude this value if the user does not provide an input for the bridge network range?

andrewtchin:
@amalashin thanks for your contribution!
@morris-jason can you review this too?

installer/engine_installer/server.go
```diff
@@ -131,6 +132,7 @@ func indexHandler(resp http.ResponseWriter, req *http.Request) {
 
 			html.PublicNetwork = getSelectOptionHTML(opts.Networks, publicNetName)
 			html.BridgeNetwork = getSelectOptionHTML(opts.Networks, bridgeNetName)
+			html.BridgeNetworkRange = getSelectOptionHTML(opts.Networks, bridgeNetRangeName)
```
morris-jason: I'm not sure this is correct. The Bridge Network Range is a CIDR range where as `opts.Networks` is going to be a list of vSphere networks.
I think the proper thing to do here is `html.BridgeNetworkRange = "172.16.0.0/12"` which is the default vic-machine range.

installer/engine_installer/engine_installer.go
morris-jason: If you make my changes below, you probably don't need to make Andrew's change as a default will always be provided.

andrewtchin:
Demo VCH Installer Wizard no longer exists, please use the H5 client plugin

Issue #791, opened 2017-09-11
Document post-upgrade housekeeping


stuclem:
Per conversation with @andrewtchin, we should document that after a successful upgrade, you should move the data disk from the previous installation into the datastore folder for the new appliance.

1. shut down the appliance, then move the disk
2. edit settings, delete Hard disk 2
3. edit settings, add Hard disk 2 from new location
4. Power on appliance

TBC if we can rename the disk too.

PR #792, opened 2017-09-11
Documented to move disk during upgrade

stuclem:
Fixes https://github.com/vmware/vic-product/issues/791


stuclem:
Thanks @andrewtchin !

PR #812, opened 2017-09-14
Adding PDF and adding missing content

stuclem:


stuclem:
Thanks @andrewtchin and @anchal-agrawal !

Issue #740, opened 2017-09-06
Document OVA deployment scenarios with multiple registries


stuclem:
I have some questions about how a user can/might/should set up multiple instances of VIC Registry.

**What I Did**

- Deployed two instances of VIC appliance (fresh installations, not upgrades) to the same vCenter Server/PSC. Let's call the appliances VIC1 and VIC2, their respective Registry instances Reg1 and Reg2, and the respective Management Portals MP1 and MP2. 
- In MP1, registered Reg2 as a replication endpoint for Reg1. Created and enabled a replication rule in the `default-project` in MP1, that uses this endpoint. I had to disable remote cert verification in MP1 for this to work.
- In MP2, registered Reg1 as a replication endpoint for Reg2. Created and enabled a replication rule in the `default-project` in MP2, that uses this endpoint. I had to disable remote cert verification in MP2 for this to work.
- Pushed a `busybox` image to Reg1. It replicated nicely to Reg2.
- Pushed an `ubuntu` image to Reg2. It replicated nicely to Reg1.
- Pushed `busybox` image to Reg2. It did not appear to replicate to Reg1, since Reg1 already has that image. 
- In addition to configuring Reg2 as a replication endpoint for Reg1, I also registered Reg2 as a source registry in MP1.

**Questions**

- [x] Do we support deploying the two OVAs, VIC1 and VIC2 to the same vCenter Server instance, i.e. registering both instances with the same PSC? Could this cause problems with users?
- [x] Is two-way replication as described above supported? Is it something users should do or avoid?
- [x] Is it always necessary to disable remote cert verification if the replication endpoint is a VIC Registry?
- [x] Do we support registering a registry, e.g. Reg2, as a source registry in MP1 if it is already a replication endpoint for Reg1? Is there any reason why users should or shouldn't do it?
- [x] If you do everything that I did above, could/should you then shut down the MP, fileserver, Demo VCH, services on VIC2, leaving only the Reg2 service running, and manage Reg2 and Reg1 exclusively through MP1? I'm thinking that if Reg2 is registered as a source registry in MP1, then you might not want MP2 to linger around for people to log into and mess about with. Afterall, in this case, the only part of VIC2 that you are interested in is Reg2.

I think that the best people to answer these questions might be @reasonerjt @andrewtchin @sergiosagu, but I decided to put it in an issue so that others can be dragged in easily. 

Thanks in advance!

cc @alextopuzov

stuclem:
[...]

PR #832, opened 2017-09-18
Removed limitation of one VIC appliance per VC

stuclem:
Also added a limitation that you can only register a VCH with one instance of MP.

Towards https://github.com/vmware/vic-product/issues/740.

@mlh78750 can you please review?


stuclem:
Thanks @mlh78750 !

Issue #849, opened 2017-09-21
Document that you must use Flex client to deploy OVA


stuclem:
We do state that you should use the vSphere Web Client, but need to make it more explicit that OVA deployment does not work with the H5 or legacy C# clients.

PR #851, opened 2017-09-21
Clarified that you must deploy OVA from Flex client

stuclem:
Fixes https://github.com/vmware/vic-product/issues/849


stuclem:
Thanks @andrewtchin

PR #855, opened 2017-09-21
Add support for upgrading OVA 1.2.x to 1.2.y

### LINKED ISSUE #839: OVA upgrade from 1.2.x to 1.2.y
This issue is for delivering an OVA upgrade procedure from 1.2.x to 1.2.y, which is different from 1.1.1 to 1.2.x since data migration for Admiral is not involved.

**Please see https://github.com/vmware/vic-product/issues/839#issuecomment-330950081 for more details. Leaving the original issue details below for posterity.**

---

/cc @sergiosagu @gmuleshkov @mdubya66 

**Data migration script:**

https://github.com/vmware/admiral/blob/master/upgrade/src/main/resources/migrate.sh

**Upgrading from:**

https://storage.googleapis.com/vic-product-ova-builds/vic-v1.2.1-rc1-15e8cde5.ova
(Post 1.2.0 build containing Admiral v1.2.1)

To test data migration, this OVA instance had (all actions performed through Admiral):

* an additional project, `project-2`
* the `busybox` image pushed to the OVA's Harbor repository for both projects
* a VCH as a Container Host with one `busybox` container running `/bin/sh` in `default-project` 

**Upgrading to:**

https://storage.googleapis.com/vic-product-ova-builds/vic-5b3261d8-devrc3.ova
(Post 1.2.1-rc2 build containing Admiral v1.2.1, Harbor 1.2.0 and Engine 1.2.1-rc2. Contains fix for OVA upgrade with external PSC: https://github.com/vmware/vic-product/pull/838)

**Observed behavior:**

Following the upgrade procedure from https://vmware.github.io/vic-product/assets/files/html/1.2/vic_vsphere_admin/upgrade_appliance.html, step 12:

```
# ./upgrade_1.1_to_1.2.sh 
Getting password from harbor.cfg
Enter vCenter Server FQDN or IP: <vCenter>
Enter vCenter Administrator Username: <user>
Enter vCenter Administrator Password: <password>
If using an external PSC, enter the FQDN of the PSC instance (leave blank otherwise): 
If using an external PSC, enter the PSC Admin Domain (leave blank otherwise): 
Starting upgrade 2017-09-19 22:50:47 +0000 UTC
Preparing upgrade environment
Disabling and stopping Admiral and Harbor path startup
Finished preparing upgrade environment
Performing pre-upgrade checks
Starting Admiral upgrade
[=] Shutting down Harbor and Admiral
[=] Upgrade Admiral
Data migration to new Admiral failed
```

**Debugging:**

The `migrate.sh` script references a `$auth_token` variable in two places:
* https://github.com/vmware/admiral/blob/master/upgrade/src/main/resources/migrate.sh#L20 and
* https://github.com/vmware/admiral/blob/master/upgrade/src/main/resources/migrate.sh#L26. 

I didn’t find where this variable was defined. This could be a typo - I see `$3` (the correct token input parameter) used in other places in this file, should we be using the same token value here?

I changed `$auth_token` to `$3` but that didn’t fix things - this could be a red herring since I'm not fully aware of how the `migrate.sh` script's environment is set up, but I think it's worth mentioning.

I then ran the `migrate.sh` script manually with the parameters it’s invoked with in the [upgrade script](https://github.com/vmware/vic-product/blob/master/installer/packer/scripts/upgrade/upgrade_1.1_to_1.2.sh):

```
# /etc/vmware/admiral/migrate.sh 10.161.67.28:8283 10.161.67.28:8282 $psc_token
shell-init: error retrieving current directory: getcwd: cannot access parent directories: No such file or directory
Old admiral node is ready for migration: 10.161.67.28:8283
New admiral node is ready for migration: 10.161.67.28:8282
Starting migration of the data
<... nothing for 5-10 mins ...>
Status code from migration: 500
Error during migration please check xenon logs for more information!
```

The script fails here: https://github.com/vmware/admiral/blob/master/upgrade/src/main/resources/migrate.sh#L59

**Summary:**

I’m not sure how to check xenon logs here. The only other thing I can think of is that **this is a 1.2.x -> 1.2.y upgrade, not a 1.1.x to 1.2.y upgrade**. Are there expectations around data schema/format during data migration? Does the migration code expect the data to follow the data schema/format from 1.1.1? In the upgrade script, I see we’re launching an Admiral 1.1.1 container to upgrade from, regardless of the version of Admiral in the older OVA (1.2.0 in this case): https://github.com/vmware/vic-product/blob/master/installer/packer/scripts/upgrade/upgrade_1.1_to_1.2.sh#L281

anchal-agrawal:
This commit modifies and renames the `upgrade_1.1_to_1.2.sh` script
to `upgrade.sh`. The script now supports upgrading from `1.2.x` to
`1.2.y` as well.

The script now detects the presence of Admiral's PSC config file
(`/data/admiral/configs/psc-config.properties`) to detect the version
of the older OVA. If this file exists on the data disk, the OVA must
be of version `1.2.x`. Then, data migration is skipped but Harbor's
address is updated in Admiral's configuration.

Fixes #839

### DIFF
--- a/installer/packer/scripts/upgrade/upgrade.sh
+++ b/installer/packer/scripts/upgrade/upgrade.sh
@@ -350,7 +350,7 @@ function upgradeAdmiral {
   if [ -n "$(docker ps -q -f name=vic-upgrade-admiral)" ]; then
     echo "Admiral upgrade container already exists" | tee /dev/fd/3
     echo "If upgrade is not already running, execute the following command and rerun the upgrade script:" | tee /dev/fd/3
-    echo "    docker rm -f vic-upgrade-admrial" | tee /dev/fd/3
+    echo "    docker rm -f vic-upgrade-admiral" | tee /dev/fd/3
     exit 1
   fi
 
@@ -580,6 +580,14 @@ function main {
   echo "-------------------------"
   echo "Starting upgrade ${TIMESTAMP}" | tee /dev/fd/3
 
+  data_upgrade_needed=false
+  if [ ! -f "/data/admiral/configs/psc-config.properties" ]; then
+      echo "Detected old OVA's version as 1.1.x. Continuing upgrade operation with data migration." | tee /dev/fd/3
+      data_upgrade_needed=true
+  else
+      echo "Detected old OVA's version as 1.2.x. Continuing upgrade operation without data migration." | tee /dev/fd/3
+  fi
+
   echo "Preparing upgrade environment" | tee /dev/fd/3
   disableServicesStart
   registerAppliance
@@ -587,9 +595,25 @@ function main {
   writeTimestamp
   echo "Finished preparing upgrade environment" | tee /dev/fd/3
 
-  upgradeAdmiral
+  if [ "$data_upgrade_needed" = true ]; then
+      upgradeAdmiral
+  else
+      # Start Admiral
+      echo "Starting Admiral" | tee /dev/fd/3
+      systemctl start admiral_startup.service
+      sleep 3
+  fi
+
   updateAdmiralConfig
-  upgradeHarbor
+
+  if [ "$data_upgrade_needed" = true ]; then
+      upgradeHarbor
+  else
+      # Start Harbor
+      echo ""
+      echo "Starting Harbor" | tee /dev/fd/3
+      systemctl start harbor_startup.service
+  fi
   setDataVersion
 
   enableServicesStart
--- a/installer/packer/scripts/upgrade/upgrade.sh
+++ b/installer/packer/scripts/upgrade/upgrade.sh
@@ -22,6 +22,8 @@ harbor_psc_token_file="/etc/vmware/psc/harbor/tokens.properties"
 admiral_psc_token_file="/etc/vmware/psc/admiral/tokens.properties"
 timestamp_file="/registration-timestamps.txt"
 
+data_upgrade_needed=false
+
 admiral_upgrade_status="/etc/vmware/admiral/upgrade_status"
 harbor_upgrade_status="/etc/vmware/harbor/upgrade_status"
 upgrade_log_file="/var/log/vmware/upgrade.log"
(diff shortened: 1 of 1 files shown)


installer/packer/scripts/upgrade/upgrade.sh
```diff
@@ -565,16 +565,40 @@ function main {
...
   updateAdmiralConfig
-  upgradeHarbor
+
+  if [ "$data_upgrade_needed" = true ]; then
+      upgradeHarbor
+  else
+      # Start Harbor
+      echo ""
+      echo "Starting Harbor" | tee /dev/fd/3
+      systemctl start harbor_startup.service
```
wy65701436: LGTM

installer/packer/scripts/upgrade/upgrade.sh
```diff
@@ -350,7 +350,7 @@ function upgradeAdmiral {
   if [ -n "$(docker ps -q -f name=vic-upgrade-admiral)" ]; then
     echo "Admiral upgrade container already exists" | tee /dev/fd/3
     echo "If upgrade is not already running, execute the following command and rerun the upgrade script:" | tee /dev/fd/3
-    echo "    docker rm -f vic-upgrade-admrial" | tee /dev/fd/3
+    echo "    docker rm -f vic-upgrade-admiral" | tee /dev/fd/3
```
andrewtchin: wow good catch

installer/packer/scripts/upgrade/upgrade.sh
```diff
@@ -565,16 +565,40 @@ function main {
   echo "-------------------------"
   echo "Starting upgrade ${TIMESTAMP}" | tee /dev/fd/3
 
+  data_upgrade_needed=false
+  if [ ! -f "/data/admiral/configs/psc-config.properties" ]; then
+      echo "Detected old OVA's version as 1.1.x. Continuing upgrade operation with data migration." | tee /dev/fd/3
```
andrewtchin: Are you adding a confirmation prompt?

anchal-agrawal: Done in 9abd7293b90c6f22c894a180c4cb049f31f8db07.

PR #939, opened 2017-10-17
Remove vic-machine-server port config

andrewtchin:
API always runs on 8443 now
Tested with deployed ova
```
chin@CORPORATE:~|⇒  curl -k https://10.160.193.206:8443/container/version
v1.2.0-rc1-0-68b5a25e%
```


installer/build/scripts/harbor/start_harbor.sh
```diff
@@ -57,4 +57,3 @@ docker rm -f dch-push
 /usr/local/bin/docker-compose -f /etc/vmware/harbor/docker-compose.yml \
                               -f /etc/vmware/harbor/docker-compose.notary.yml \
                               -f /etc/vmware/harbor/docker-compose.clair.yml up
-
```
zjs: It doesn't seem like this was an intentional change.

andrewtchin: it was, just removing empty line i saw

PR #954, opened 2017-10-18
VIC Appliance design doc

andrewtchin:
Fixes #880

PR #960, opened 2017-10-19
Separate persistent disks

### LINKED ISSUE #743: Separate persistent data to individual disks on the vSphere Integrated Containers Appliance
Tracking for separating the MySQL DB for the registry and management portal to second vDisk and separate the log files to a third vDisk.

Use case to mitigate issues if the root disk becomes full.

Bug Ref# 1954150

morris-jason:
Fixes #743, consistent with storage design proposals in #954.

Also includes a hot fix for failing fast in the cache script if a curl request returns an error.
Adds the `-it` flag to building the docker container locally.


installer/build/baseimage/stage.sh
```diff
@@ -320,47 +318,52 @@ done
...
-convert "$dev2" "$DATAROOT" "$IMG2NAME" "$VMDK2NAME"
+section "export images to VMDKs"
+for i in "${!IMAGES[@]}"; do
+  progress "exporting ${IMAGES[$i]}.img to ${IMAGES[$i]}.vmdk"
+  convert "${DEVS[$i]}" "${IMAGEROOTS[$i]}" "${IMAGES[$i]}.img" "${IMAGES[$i]}.vmdk"
+done
 
 task "VMDK Sizes"
-du -h bin/*
+du -h bin/*
```
andrewtchin: is this used to set the size in the ovf at the end? i think that's not happening correctly on master

installer/build/scripts/provision_harbor.sh
```diff
@@ -38,7 +38,11 @@ for _, s in enumerate(dataMap["services"]):
         dataMap["services"][s]["restart"] = "on-failure"
   if "volumes" in dataMap["services"][s]:
     for kvol, vol in enumerate(dataMap["services"][s]["volumes"]):
-      if vol.startswith( '/data' ):
+      if vol.startswith( '/data/database' ):
```
andrewtchin: can you put a comment to note that this is fixing up the compose file?

installer/build/scripts/systemd/engine_installer/engine_installer.service
```diff
@@ -1,8 +1,8 @@
 [Unit]
 Description=Demo VCH Installer Wizard Service
 Documentation=https://github.com/vmware/vic
-After=engine_installer_startup.service data.mount
-Requires=data.mount
+After=engine_installer_startup.service .targetvic-mounts.target
```
andrewtchin: is this a typo?

andrewtchin:
reminder to ping stuart to update docs and system requirements for this

installer/build/baseimage/stage.sh
morris-jason: No, it just prints out the value. ovf vmdk file reference sizes are definitely not correct in master, and haven't been for a long time.

installer/build/scripts/provision_harbor.sh
morris-jason: sure.

installer/build/scripts/systemd/engine_installer/engine_installer.service
morris-jason: yes, good catch!

andrewtchin:
Remove this since it's not needed for upgrade anymore https://github.com/vmware/vic-product/blob/master/installer/build/cache.sh#L31

andrewtchin:
Also if logs will be lost during the upgrade to new disks, print a warning in the upgrade script with a confirm y/n input

installer/build/scripts/upgrade/util.sh
```diff
@@ -0,0 +1,44 @@
+#!/usr/bin/env bash
```
andrewtchin: nice

PR #968, opened 2017-10-20
Update Demo VCH Installer Wizard

### LINKED ISSUE #890: Description of Demo VCH Wizard Port option in OVA installer says "fileserver"
**VIC Product version:**

`vic-v1.2.1-rc3-4104e5f9.ova`

**Deployment details:**

VC on Windows, embedded PSC.

**Steps to reproduce:**

During OVA deployment, the description of step 6.1 **Engine Install Wizard Port** includes "fileserver":

wizfileserver

Also, "Engine" in step 6.1 should be "Demo VCH".

**Actual behavior:**

The description in step 6.1 is a copy/paste of step 5.1.

**Expected behavior:**

Step 6.1 should be something like 

**6.1 Demo VCH Install Wizard Port -- Specifies the port on which the installer will be published.**

### LINKED ISSUE #895: Update Demo VCH Installer Wizard
**User Statement:**

As a user I want this fixed.

**Details:**
Bugzilla 1957067

**Acceptance Criteria:**
- [ ] Implement fix

andrewtchin:
Fixes #895 
Fixes #890

### DIFF
--- a/installer/engine_installer/html/exec.html
+++ b/installer/engine_installer/html/exec.html
@@ -39,7 +39,7 @@
         $.ajax({
         	    type: "post", url: "/cmd", contentType: "application/json", data: JSON.stringify(data),
         	    success: function(data) {
-                $("#create").text(data).html()
+                $("#create").text(data)
                 if (window["WebSocket"]) {
                   ws = new WebSocket("wss://" + document.location.host + "/ws");
                   ws.onmessage = function (msg) {
--- a/installer/engine_installer/html/exec.html
+++ b/installer/engine_installer/html/exec.html
@@ -20,7 +20,7 @@
     $( document ).ready(function() {
       var ws;
       function appendLog(item) {
-        $("#log").append("<h6>"+item+"</h6>")
+        $("#log").append(document.createTextNode(item))
       }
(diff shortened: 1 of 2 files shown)


installer/engine_installer/html/exec.html
```diff
@@ -38,7 +39,7 @@
         $.ajax({
         	    type: "post", url: "/cmd", contentType: "application/json", data: JSON.stringify(data),
         	    success: function(data) {
-                $("#create").html("<p>"+data+"</p>")
+                $("#create").text(data).html()
```
zjs: What is the purpose of the `.html()` on this line?

andrewtchin: I'm not sure @pstroia gave me that example. It seems to work the same either way

zjs: My jQuery is a bit rusty, but I believe `.html()` (no arguments) acts as an accessor for the contents of the selected element. Given that we do nothing with the return value, I suspect it could be removed.

andrewtchin: you're right, confirmed with josh. updated now

pstroia:
Ok.

PR #975, opened 2017-10-23
Compose file without volumes

pescerosso:


mdubya66:
@andrewtchin do these need tests?

andrewtchin:
These are for demos/tutorials and not included in our builds so I'll leave that up to the author

PR #1037, opened 2017-11-03
Vendor vmware/vic

### LINKED ISSUE #353: Fix VIC inclusion strategy and ci ova biuld
Opening this in response to #352.

### VIC Inclusion

Some of the installer/engine_installer code has dependencies on vic-engine, such as vic/pkg/certificate. This may increase in the future as we refactor the install to use vic-engine as a library instead of invoking exec commands on the vic-binary.

The simplest idea is to clone the vic repo to the ci container during drone builds.
```bash
# Grab the dependencies from vic if we're running ci
if [[ $DRONE ]]; then
  mkdir -p /go/src/github.com/vmware/vic/
  git clone https://github.com/vmware/vic.git /go/src/github.com/vmware/vic/
fi
```

In the future we may want to vendor the vic repo - that would be in line with the idea of the ova packaging up a specific version of the vic products. 
However, this decreases the ability make rapid changes to the ova if we have to wait on release versions of vic-engine.

It's possible we can do hybrid strategy like we currently do with Admiral and Harbor. 
Release versions of the OVA would be built with numbered (ie. 1.1.1) versions of all components using repo tags, and dev builds are pulling down the most recent containers for those products.

### OVA Build Concerns

There is also an issue with building the ova in the ci using the current build script. The build script is running in the integration-ci container, but needs SSH access to the PACKER ESX host. We need to investigate if drone can give port access to agent containers or if we need an alternative packer builder for the ci.

morris-jason:
Fixes #353. Vendors vmware/vic into vic-product for use in the fileserver binary.

Switches `gvt` to [`dep`](https://github.com/golang/dep) which flattens vendor direcories automatically. It also provides a huge decrease in repo size compared to gvt, about 100mb less.


mdubya66:
LGTM

PR #1043, opened 2017-11-07
[skip ci]Move vic product tests from vic repo to vic-product repo

mhagen-vmware:
Most of these tests need to be tweaked/re-written still, but this at least gets it moving in the right direction


mdharamadas1:
@mhagen-vmware what is the plan here? So we are just moving tests as is without keywords migrated and without making sure they pass on VC? do you plan to delete them from vic repo? 
Instead, why not move them completely , making sure they pass on CI before merging to master? Also, would be great we setup to run these tests in parallel before moving them from vic.

mhagen-vmware:
1. I already have a corresponding PR to delete the tests from vic repo
2. There is a mix of incomplete tests, and not applicable tests anymore. This is a first step to at least get them in the correct place.  Afterwards we will need to start working on cleaning them up.
3. I would rather have them at least in the correct location sooner rather than later, than spending days and days getting tests that in some cases never worked to work

mdharamadas1:
OK, if they are not running then makes sense. But these look like nightly tests, shouldn't they go under some other dir like `manual-test-cases`? Currently, CI is set up to run all tests under `tests/test-cases` by default.

mhagen-vmware:
ah, you are right - didn't realize you were running all of test cases folder already.  move to manual now

PR #1050, opened 2017-11-08
Tutorial - Custom Certificates for VCHs

emeirell:
Tutorial on how to create custom certificate for VCH

### DIFF
--- a/tutorials/custom_certificate/README.md
+++ b/tutorials/custom_certificate/README.md
@@ -13,15 +13,14 @@ The certificate requirements for VCH are:
     
 **Create an RSA Private Key**
 
-`openssl genrsa -des3 -out fullkey.key 2048 pkcs8`
-    
-**Remove passphrase from key**
-
-`openssl rsa -in fullkey.key -out mykey.key`
-    
+`openssl genrsa -out mykey.key 2048 pkcs8`
+     
 **Generate CSR**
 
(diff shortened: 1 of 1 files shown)


tutorials/custom_certificate/README.md
```diff
@@ -0,0 +1,29 @@
...
+The certificate requirements for VCH are: 
+* an X.509 certificate
+* KeyEncipherment 
+* DigitalSignature 
+* KeyAgreement 
+* ServerAuth
+    
+**Create an RSA Private Key**
+
+`openssl genrsa -des3 -out fullkey.key 2048 pkcs8`
```
andrewtchin: If you don't specify `-des3`, the private key is not encrypted and will not require the below step. I would recommend changing this to `openssl genrsa -out key.pem 4096`

tutorials/custom_certificate/README.md
```diff
@@ -0,0 +1,29 @@
...
+    
+**Remove passphrase from key**
+
+`openssl rsa -in fullkey.key -out mykey.key`
+    
+**Generate CSR**
+
+`openssl req -new -key mykey.key -out server.csr -config template.cnf`
```
andrewtchin: `template.cnf` isn't specified anywhere, so this command fails. I think this config file should be included in the tutorial or specified as parameters to the command if possible.

PR #1055, opened 2017-11-09
Support harbor upgrade 1.2 -> 1.3

wy65701436:


installer/build/scripts/upgrade/upgrade-harbor.sh
```diff
@@ -128,23 +144,34 @@ function configureHarborCfgManageKey {
...
-  # Add generated clair_db_password as managed key if not present
-  configureHarborCfgUnset clair_db_password "$(genPass)" true
+  # Add generated log_rotate_days, log_rotate_size, db_host, db_port, db_user, uaa_endpoint, 
+  # uaa_clientid, uaa_clientsecret and uaa_ca_root as managed key if not present
+  configureHarborCfgUnset log_rotate_days 3  
+  configureHarborCfgUnset log_rotate_size 200M
+  configureHarborCfgUnset db_host mysql
+  configureHarborCfgUnset db_port 3306
+  configureHarborCfgUnset db_user root 
+  configureHarborCfgUnset uaa_endpoint uaa.mydomain.org
```
andrewtchin: what are these used for and are these changed by harbor startup or something later? i'm wondering if harbor starts ok with these default values

wy65701436: The uaa part will not impact VIC even it's changed, it's for PKS integration only. 
And it's ok to launch Harbor with all the default values.

PR #1068, opened 2017-11-13
Create vic-appliance-ready.target

andrewtchin:


installer/build/scripts/systemd/admiral/admiral_startup.service
```diff
@@ -1,8 +1,8 @@
 [Unit]
 Description=Admiral Startup Configuration
 Documentation=http://github.com/vmware/admiral
-After=docker.service vic-mounts.target systemd-resolved.service iptables.service ova-firewall.service admiral_startup.path firstboot.path
-Requires=iptables.service vic-mounts.target
+After=vic-appliance-ready.target admiral_startup.path firstboot.path
```
morris-jason: Can we remove firstboot.path here?

installer/build/scripts/systemd/vic-appliance.target
```diff
@@ -1,6 +1,6 @@
 [Unit]
-Description=System customization for VIC(Integrated containers engine on vSphere) based on user provided parameters
-Documentation=https://github.com/vmware/vic/blob/master/doc/bundle/OVA-spec.md
+Description=System customization for VIC Appliance
+Documentation=https://github.com/vmware/vic-product/blob/master/installer/DESIGN.md
 Requires=vic-mounts.target firstboot.service
```
morris-jason: Will requiring `firstboot.service` here will trigger `firstboot.sh` any time `vic-appliance.target` is required?
It may be wise to require on `firstboot.path` here instead. (Meaning we may need to rename `firstboot.{sh,service}` to something like `load-docker-containers`.

installer/build/scripts/systemd/admiral/admiral_startup.service
andrewtchin: fixed

andrewtchin:
Confirmed restarting the appliance all services came up

Issue #1075, opened 2017-11-14
Update how to access VIC view in vSphere Client


stuclem:
Per @lgayatri, the doc gives incorrect info about how to access VIC in the vSphere Client:
https://vmware.github.io/vic-product/assets/files/html/1.2/vic_vsphere_admin/access_h5_ui.html says click Home, but from vSphere 6.5u1 onwards, the link is under Shortcuts

PR #1078, opened 2017-11-15
Fixed path to VIC in vSphere Client

stuclem:
Fixes https://github.com/vmware/vic-product/issues/1075.

@lgayatri and @jooskim is this OK? Do either of you know if the Shortcuts view was in in the h5 client in 6.5.0, or whether it was introduced in 6.5u1? I have a suspicion that in 6.5.0 the shortcut icons appeared in Home, but maybe I'm wrong. Thanks!


stuclem:
Also fixed one instance of the path to VCHs in Management Portal, which still referred to clusters rather than to container hosts

stuclem:
Confirmed that in 6.5.0 the VIC icon is indeed in the Home page, not in the Shortcuts page.

b0c0fc6d

However, in both 6.5.0 and 6.5u1, clicking on the vSphere Client logo sends you to the correct place. Updated the doc accordingly in https://github.com/vmware/vic-product/pull/1078/commits/1be7010bbe75353f593fd42ee2280abc22053ae7.

alextopuzov:
LGTM.

Issue #961, opened 2017-10-19
TS doc ovf environment


andrewtchin:
If when booting the VIC Appliance the console screen displays the error `unable to unmarshal ovf environment`, the appliance will be in an unusable state.

Solution:
- Power off guest OS
- Right click the appliance, Edit Settings
- Go to `vApp Options`, click OK
- Verify that a `Reconfigure virtual machine` task is completed for the appliance
- Power on the appliance
- Verify that the console does not display error message

These steps rewrite a corrupted or missing ovfenv data.

andrewtchin:
related https://github.com/vmware/vic-product/issues/741

anba89:
Hi 

My VIC Appliance displays the error unable to unmarshal ovf environment and I tried the steps above without success. Is there another solution?

stuclem:
Thanks for raising this @anba89. @andrewtchin do you have any ideas?

andrewtchin:
@anba89 Sometimes it takes multiple attempts following the entire procedure above to write the OVF data again. Please try a few times and let us know if it doesn't work still.

anba89:
@andrewtchin I tried it more than six times and then I redeployed the appliance.

andrewtchin:
Thanks @anba89 could you provide a description of the environment you deployed in? vCenter version, datacenter configuration, datastore configuration, etc? I will get an issue opened in vCenter for this.

anba89:
Hi @andrewtchin 

I simulated an ESXi crash today. And was able to restore the new VIC appliance with the documented steps. 

vCenter version: 6.5.0, 7119157
ESXi version: 6.5.0, 5224529
Datastore: NFSv3
Environment: Cisco HyperFlex Solution

stuclem:
Thanks @anba89. Can we close this one again now?

andrewtchin:
Yes closing.

PR #1085, opened 2017-11-16
Added OVF TS topic

stuclem:
Fixes https://github.com/vmware/vic-product/issues/961.

@andrewtchin, in https://github.com/vmware/vic-product/issues/961 you specify **Edit Settings** > **vApp Options**. As far as I can see this only appears in the Flex client. In H5, there is only a **VM Options** tab. Would going here have the same effect?


andrewtchin:
I haven't tried the H5C - as long as it shows the Reconfigure task after you use that option I think it should be the same

stuclem:
@andrewtchin the Reconfigure task does appear if you select VM Options in the H5 client, so I think that we're OK.

Issue #945, opened 2017-10-17
Bridge network explanation change request. 


dbarkelew:
https://vmware.github.io/vic-product/assets/files/html/1.2/vic_vsphere_admin/vch_networking.html

The requirements does not make it clear what the VLAN is for resulting in using the same DVS port group and VLAN ID for multiple VCH deployments. 

Suggest explaining that the bridge has its own DHCP service that needs to be on an isolated VLAN that is trunked between the ESXi hosts in the cluster that the VCH will be deployed to. Also explain to use Virtual Switch Tagging for the VLAN ID and that a new VLAN ID and DVS port group is needed for each additional VCH.

See KB on VST
* VLAN configuration on virtual switches, physical switches, and virtual machines
http://kb.vmware.com/kb/1003806

@pescerosso

dbarkelew:
The idea is to have the requirement clear enough for any network admin to understand the request and provide the needed changes. Essentially need to say VLANs are needed for the bridge so the VCH can provide the containers DHCP services without conflicts with other DHCP services the datacenter may have elsewhere. Even NSX VNIs (virtual wires) can be used to achieve the same.

stuclem:
Comments from @pescerosso via Slack:

-----

https://vmware.github.io/vic-product/assets/files/html/1.2/vic_vsphere_admin/vic_installation_prereqs.html#networkreqs
In particular this
the information is there, I guess 
but it is very difficult to parse I think
Wouldn’t it be more clear if we have a paragraph with prerequisite for the bridge network, maybe linking to an exist clear network config that VI admin already know
So that I can cut and paste and pass it to the network

------

PR #1086, opened 2017-11-16
Updated bridge network docs

stuclem:
Fixes https://github.com/vmware/vic-product/issues/945 and incorporates information received via Slack and email from @pdaigle and @pescerosso. Attempts to homogenize the bridge network requirements in both the topic on bridge networks, and in the VCH deployment prereqs.

@dbarkelew @pdaigle @pescerosso can you please take a look? I would appreciate if you could look at [Configure Bridge Networks](https://github.com/stuclem/vic-product/blob/bridgevlan/docs/user_doc/vic_vsphere_admin/bridge_network.md) in its entirety. Thanks!


stuclem:
@pdaigle another question: everywhere that I mention vSphere Distributed Switch, should I also be mentioning NSX Virtual Switch?

stuclem:
OK, thanks @tschwaller for the clarification

PR #1101, opened 2017-11-17
Populate ManagedByInfo in OVA

jzt:
This change populates the `ManagedBy` field of the OVA so that it can be detected/managed by the plugin. Fixes backend side for https://github.com/vmware/vic-ui/issues/26.

Will work side-by-side with https://github.com/vmware/vic-ui/issues/83.

### DIFF
--- a/installer/build/scripts/admiral/configure_admiral.sh
+++ b/installer/build/scripts/admiral/configure_admiral.sh
@@ -157,10 +157,6 @@ function secure {
 function detectHostname {
   hostname=$(hostnamectl status --static) || true
   if [ -n "$hostname" ]; then
-    if [ "$hostname" = "localhost.localdomain" ]; then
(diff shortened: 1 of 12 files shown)


installer/configure/configure.go
```diff
@@ -0,0 +1,76 @@
...
+	return nil
+}
+
+// Run takes in a url and session and sets the MangedBy field with the vic-machine HTML5 extension package name.
+func Run(ctx context.Context, sess *session.Session) error {
+	vm, err := guest.GetSelf(ctx, sess)
+	if err != nil {
+		return errors.Errorf("failed to get product vm : %s", errors.ErrorStack(err))
+	}
+
```
andrewtchin: Maybe add a log saying we're starting the attempt to set this

PR #1103, opened 2017-11-19
[WIP] Fix startup timing of components

### LINKED ISSUE #891: Docker client fails authorization after OVA reset
When I install the OVA, I go through the various processes to add the root certificate of the registry to a remote Docker and then I can login fine using my PSC vSphere ID. Eg.

```
vagrant@ubuntu-1604-vmware:~$ docker login 10.118.68.183
Username: bcorrie@vsphere.local
Password: 
Login Succeeded
```
I can do this without adding this ID anywhere in the Management UI. I don't need to be a member of a project or have any kind of role.

If I reset (power on / off) the OVA, I can no longer successfully login. I get the following message:

```
vagrant@ubuntu-1604-vmware:~$ docker login 10.118.68.183
Username: bcorrie@vsphere.local
Password: 
Error response from daemon: login attempt to https://10.118.68.183/v2/ failed with status: 401 Unauthorized
```
What's interesting is that the authentication appears to be working fine, but it fails authorization for some unknown reason. Even if I add this ID as an Admin to the default project, as a top level management Admin, to a new project. Doesn't matter. It still won't authorize me.

I tried re-downloading the root certificate thinking maybe it was regenerated. It was the same as the original one and didn't help.

The error I'm seeing in the Harbor logs is

```
Sep 27 17:36:16 172.18.0.1 registry[367]: time="2017-09-27T17:36:16.828684058Z" level=info msg="token signed by untrusted key with ID: \"6T3F:EYCE:II65:PYW7:6V3D:ZW57:H77J:J4RC:4A7U:ISI5:KST2:HEKE\""
Sep 27 17:36:16 172.18.0.1 registry[367]: time="2017-09-27T17:36:16.828813392Z" level=warning msg="error authorizing context: invalid token" go.version=go1.7.3 http.request.host=10.118.68.183 http.request.id=9ccce67b-22a7-46e2-b846-227c9784634c http.request.method=GET http.request.remoteaddr=10.118.69.174 http.request.uri="/v2/" http.request.useragent="docker/17.03.0-ce go/go1.7.5 git-commit/60ccb22 kernel/4.4.0-31-generic os/linux arch/amd64 UpstreamClient(Docker-Client/17.03.0-ce \\(linux\\))" instance.id=453a5f82-8977-4d66-94ed-f2c87f9aaaf3 service=registry version=v2.6.2
```
So what we need to figure out is:

1 - What is happening on boot that is causing this authorization problem?
2 - Does this mean that after a reboot, no user is able to pull or push images from the Registry? If so, then this is a ship stopper for 1.2.1 IMHO.

morris-jason:
~Fixes #891 with a cleanup of systemd services.~
I'm opening a separate PR to address 891, and this is more of a systemd overhaul for the appliance. Not needed for 1.3.

The approach here is to simplify component configuration by moving the configure scripts to a ExecStartPre step of the running component service unit.
I've added a `vic-components` target to give admins an easier way to diagnose if a failure is from a component, the system, or an appliance service (like disk mounts).
Also ensured path files were attached to the `paths.target` runlevel.

A major problem with path files is that they are considered active on startup (albeit waiting), so service units that require them won't actually wait for the file to be present. I've added a file waiting bash script to solve this issue in component service units, but I'm also leaving the path files present for now.


morris-jason:
Closing, underlying issues fixed in #1104.

PR #1107, opened 2017-11-21
Update for new vch logs

mdharamadas1:
Changes due to default logrus format in VCH.


mdharamadas1:
@zjs yeah we don't have any way to share keywords between repos yet.

PR #1108, opened 2017-11-21
Include build number in product version

jzt:
This change includes the VIC build number in the version string.

Fixes https://github.com/vmware/vic-ui/issues/85

### DIFF
--- a/installer/build/build.sh
+++ b/installer/build/build.sh
@@ -16,7 +16,7 @@
 # exit on failure and configure debug, include util functions
 set -eu -o pipefail +h
 
-DRONE_BUILD_NUMBER=${DRONE_BUILD_NUMBER:-}
+DRONE_BUILD_NUMBER=${DRONE_BUILD_NUMBER:-0}
 export BUILD_NUMBER=${DRONE_BUILD_NUMBER:-}
 ADMIRAL=""
 VICENGINE=""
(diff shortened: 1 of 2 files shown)


installer/build/build.sh
```diff
@@ -66,8 +66,9 @@ function cleanup() {
 
 trap cleanup EXIT
 
-GIT_TAG="$(git describe --tags)"
-export BUILD_OVA_REVISION=${GIT_TAG}
+TAG=$(git for-each-ref --format="%(refname:short)" --sort=-authordate --count=1 refs/tags) # e.g. `v0.9.0`
```
zjs: This is... *weird*... logic, and we should fix it at some point, but it looks like it's already present in other repositories so I guess it's fine for now. That said, please file an issue to fix this post-1.3.

In case it's not clear what's wrong:
```
$ git log --all --graph --oneline
* 7c69e9e (tag: v1.0.1, bugfix) Fix a bug
| * c6e5d48 (HEAD -> master) Even more work
| * 9ec43d5 More work
| * f66eb7a (tag: v2.0.0) Thursday
| * 5f48f72 Wednesday
| * 2200486 Tuesday
|/  
* cea4158 Monday
* bd0efb1 (tag: v1.0.0) Initial Commit
$ git for-each-ref --format="%(refname:short)" --sort=-authordate --count=1 refs/tags
v1.0.1
```

Just because `v1.0.1` is the most recently added tag doesn't mean it's an appropriate tag to use for a build; in this hypothetical, we've done a `v1.0.1` bugfix and now it shows up as the `$TAG` for `master`.

PR #1114, opened 2017-11-21
Removed --container-store option from docs

stuclem:
Fixes https://github.com/vmware/vic-product/issues/102.

@AngieCris can you please review this? I did a global search of the docs and this was the only place in which `--container-store` was mentioned.


stuclem:
Thanks @AngieCris !

Issue #1106, opened 2017-11-20
Document that demo VCH wizard is deprecated in 1.3


stuclem:
Per Slack chat with @andrewtchin we are not adding any new features to the demo VCH wizard, but it will still be present in 1.3. The docs should mention that it's deprecated and direct users to the vSphere Client VCH deployment wizard.

stuclem:
Actually, following conversation with @mdubya66 we might remove the demo VCH wizard completely in 1.3. Leaving this one in Verify until a decision is made.

stuclem:
In the end I removed the topic on the Demo installer completely. Because the demo installer requires you to have downloaded the vic-machine bundle and run vic-machine update firewall, keeping the demo installer procedure in the doc really screws up the order that I have to present the topics in. Without it, I can present the topics in a more logical sequence.

stuclem:
Leaving this one in verify pending decision to remove in https://github.com/vmware/vic-product/issues/1144. The only remaining mentions in the doc are in the OVA installer topic and ports reference.

stuclem:
@andrewtchin @mdubya66 did we decide to leave the demo VCH wizard in the product, and just remove it from the Getting Started page?

andrewtchin:
I think we're only removing it from the Getting Started page

stuclem:
Confirmed by @mdubya66. The demo VCH wizard is marked as deprecated in the only places where it is still necessary to mention it (OVA deployment and ports ref).

PR #1115, opened 2017-11-21
Flagged demo VCH deployment wizard as deprecated.

stuclem:
Towards https://github.com/vmware/vic-product/issues/1106, pending a decision about whether or not to remove this wizard completely.

@andrewtchin and @morris-jason can you please review?


stuclem:
Thanks @andrewtchin and @morris-jason. I think that we're waiting for @pdaigle to come back from vacation to decide what to do with this installer. But in the meantime, marking as deprecated should do the trick.

Issue #749, opened 2017-09-07
Document that VCH can have one interface per network


stuclem:
https://github.com/vmware/vic/issues/2802 is fixed in PR https://github.com/vmware/vic/issues/6201. So, we can remove all of the notes about 2 networks having to share an interface from the docs.

stuclem:
This fix is definitely in 1.3. Awaiting confirmation from @andrewtchin whether it would also be in an eventual 1.2.2 release.

andrewtchin:
1.2.2 is not currently planned so determination to pull this fix in would be made if we have that release

PR #1116, opened 2017-11-21
Removed limitation that at least 2 networks must share interface

stuclem:
Towards https://github.com/vmware/vic-product/issues/749, but need to verify that no other instances of this limitation are present.


stuclem:
Removing @emlin and adding @jzt and @cgtexmex as reviewers

stuclem:
Thanks @jzt and @hickeng !

PR #1118, opened 2017-11-21
Add drone git hash check

morris-jason:


morris-jason:
Closing. Will be fixed in next version of drone.

Issue #1129, opened 2017-11-22
Document that Getting Started page takes a few minutes to appear after 1st boot


stuclem:
Per https://github.com/vmware/vic-product/pull/1068#issuecomment-344036816. New behaviour, high priority.

Issue #1098, opened 2017-11-17
Document that you can set host name in OVA deployment


stuclem:
Per https://github.com/vmware/vic-product/issues/1089 and PR https://github.com/vmware/vic-product/pull/1066. According to @morris-jason via Slack " If the user enters a hostname fqdn in the ovf, the services on the ova will be served from that fqdn, including PSC registration and admiral/harbor."

stuclem:
New feature. High priority.

PR #1141, opened 2017-11-23
Added FQDN and waiting time to OVA procedure

stuclem:
- Fixes https://github.com/vmware/vic-product/issues/1098
- Fixes https://github.com/vmware/vic-product/issues/1129

@andrewtchin and @morris-jason can you please review?

### DIFF
--- a/docs/user_doc/vic_vsphere_admin/deploy_vic_appliance.md
+++ b/docs/user_doc/vic_vsphere_admin/deploy_vic_appliance.md
@@ -39,9 +39,9 @@ You install vSphere Integrated Containers by deploying a virtual appliance. The 
 
     **IMPORTANT**: You require SSH access to the vSphere Integrated Containers appliance to perform upgrades. You can also use SSH access in exceptional cases that you cannot handle through standard remote management or CLI tools. Only use SSH to access the appliance when instructed to do so in the documentation, or under the guidance of VMware GSS.
 
-5. Expand **Networking Properties** and optionally configure a static IP address for the appliance VM. 
+5. Expand **Networking Properties** and optionally configure a static IP address and fully qualified domain name (FQDN) for the appliance VM. 
 
-    To use DHCP, leave the networking properties blank.
+    To use DHCP, leave the networking properties blank. If you specify an FQDN, the appliance uses this FQDN to register with the Platform Services Controller and runs the Registry, Management Portal, and file server services at that FQDN.
 
     **IMPORTANT**: If you set a static IP address for the appliance, use spaces to separate DNS servers. Do not use comma separation for DNS servers. 
 
@@ -89,7 +89,7 @@ You install vSphere Integrated Containers by deploying a virtual appliance. The 
 
     The network status shows whether the network settings that you provided during the deployment match the settings with which the appliance is running. If there are mismatches, power off the appliance and select **Edit Settings** > **vApp Options** to correct the network settings.
     
-11. In a browser, go to  http://<i>vic_appliance_address</i> and when prompted, enter the connection details for the vCenter Server instance on which you deployed the appliance.
+11. Wait for a few minutes to allow the appliance services to start, then in a browser, go to http://<i>vic_appliance_address</i> and enter the connection details for the vCenter Server instance on which you deployed the appliance.
 
      - The address and single sign-on credentials of vCenter Server.
      - If vCenter Server is managed by an external Platform Services Controller, enter the FQDN and administrator domain for the Platform Services Controller. If vCenter Server is managed by an embedded Platform Services Controller, leave the External PSC text boxes empty.


docs/user_doc/vic_vsphere_admin/vic_installation_prereqs.md
```diff
@@ -1,30 +1,34 @@
...
-|Mac OS X |10.11 (El Capitan)|
-|Linux|Ubuntu 16.04 LTS|
+You deploy the vSphere Integrated Containers appliance on a vCenter Server instance. Deploying the appliance directly on an ESXi host is not supported.
 
-Other recent 64-bit OS versions should work but are untested.
+- vCenter Server 6.0 or 6.5
+- ESXi 6.0 or 6.5 for all hosts
+- At least 2 vCPUs
+- At least 8GB RAM
+- At least 80GB free disk space on the datastore
```
morris-jason: Can we specify here that the appliance disk space is thin provisioned?

docs/user_doc/vic_vsphere_admin/deploy_vic_appliance.md
```diff
@@ -100,24 +89,22 @@ You install vSphere Integrated Containers by deploying a virtual appliance. The
...
+- If you see the error `Failed to register with PSC. Please check the vSphere user domain PSC settings and try again`, see the procedure in [vSphere Integrated Containers Appliance Fails to Register with PSC](ts_register_psc_fails.md) to register vSphere Integrated Containers with the Platform Services Controller.
 - If you see a certificate error when you attempt to go to http://<i>vic_appliance_address</i>, see [Browser Rejects Certificates with `ERR_CERT_INVALID` Error](ts_cert_error.md).
-
-- If necessary, you can reconfigure the appliance after deployment by editing the settings of the appliance VM. For information about reconfiguring the appliance, see [Reconfigure the vSphere Integrated Containers Appliance](reconfigure_appliance.md).   
+- If the initialization fails with any other error, see [Reinitialize the vSphere Integrated Containers Appliance](reinitialize_appliance.md). You should not reinitialize the appliance in any circumstances other than those described in that topic.
```
andrewtchin: Also it might be useful to note here that if the user does not receive a green success bar, the appliance is not correctly initialized. Some users have assumed that since the red bar for a failed install goes away when they reload the page everything is OK, but it isn't. We have an issue open to make the failure message persist.

docs/user_doc/vic_vsphere_admin/vic_installation_prereqs.md
stuclem: Thanks @morris-jason. Updated to "At least 80GB free disk space on the datastore. The disk space for the appliance uses thin provisioning."

docs/user_doc/vic_vsphere_admin/deploy_vic_appliance.md
stuclem: Thanks @andrewtchin. Merged your comment into the existing bullet about reinitializing:

"If you do not see a green success banner at the top of the Getting Started page after initializing the appliance, the appliance has not initialized correctly. For more information, see [Reinitialize the vSphere Integrated Containers Appliance](reinitialize_appliance.md). You should not reinitialize the appliance in any circumstances other than those described in that topic."

stuclem:
Another question for @andrewtchin and @morris-jason: We link to the topic [vSphere Integrated Containers Appliance Fails to Register with PSC](https://vmware.github.io/vic-product/assets/files/html/1.3/vic_vsphere_admin/ts_register_psc_fails.html) from the Troubleshooting section of the deploy OVA topic. Should we remove this from 1.3? Isn't it very 1.2.0 specific, and wasn't this resolved in 1.2.1? Thanks!

docs/user_doc/vic_vsphere_admin/vic_installation_prereqs.md
stuclem: @morris-jason does this mean that whatever you select in the OVA deployment wizard is ignored?

thin

stuclem:
@andrewtchin and @morris-jason another question about OVA deployment: It is possible to change the default ports for all VIC services, but not, it seems for the `vic-machine` server (default 8443). Is this deliberate?

docs/user_doc/vic_vsphere_admin/vic_installation_prereqs.md
morris-jason: It's my understanding that thin provisioning is the default, but that may be wrong.
If it is the default, my thinking is that if we say "at least 80GB" then people who don't have that much space may be turned away from deploying an ova. In reality the thin provisioning default takes about ~8-12gb, but that's a fluid number based on component sizes.

andrewtchin:
1) I think it would be cool to rename it and note some scenarios when manual registration would be used
2) This is deliberate as the UI plugin needs a stable port to know where to find the API

stuclem:
Thanks @andrewtchin I created https://github.com/vmware/vic-product/issues/1162 for the PSC topic.

Issue #1127, opened 2017-11-22
Remove instruction to restart flex client after upgrade


stuclem:
As pointed out by @lgayatri, if you upgrade the client plugin, only the h5 client is upgraded. So, the step to restart the flex client service is unnecessary.

PR #1142, opened 2017-11-23
Removed the steps to restart Flex clients, updated versions.

stuclem:
Fixes https://github.com/vmware/vic-product/issues/1127. 

@lgayatri and @jooskim can you please review? You can see the full versions of the upgraded topics here:

- [Windows](https://github.com/stuclem/vic-product/blob/flexupgrade/docs/user_doc/vic_vsphere_admin/upgrade_h5_plugin_windows.md)
- [VCSA](https://github.com/stuclem/vic-product/blob/flexupgrade/docs/user_doc/vic_vsphere_admin/upgrade_h5_plugin_vcsa.md)

### DIFF
--- a/docs/user_doc/vic_vsphere_admin/upgrade_h5_plugin_vcsa.md
+++ b/docs/user_doc/vic_vsphere_admin/upgrade_h5_plugin_vcsa.md
@@ -1,4 +1,4 @@
-# Upgrade the Plug-In on vCenter Server Appliance #
+# Upgrade the HTML5 Plug-In on vCenter Server Appliance #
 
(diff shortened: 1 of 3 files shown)


stuclem:
Thanks @lgayatri, fixed in https://github.com/vmware/vic-product/pull/1142/commits/a5bd57e7d4ff3043ef6bb208905b9c0ae0ec387c. Usually, it doesn't matter what numbers you use in the Markdown source for step numbers - whatever number you use in the source, the steps usually render with the correct numbering in output. But, it seems that the first number in a sequence is respected, so if your first step is numbered `3.`  in the markdown, then the output will start at 3 too. So, I have learned something new about Markdown :)

stuclem:
Thanks @jooskim. I addressed your comments in https://github.com/vmware/vic-product/pull/1192.

PR #1161, opened 2017-11-28
Remove custom attribute tag

jzt:
This removes the `vic-ova-identifier` custom attribute from the OVA.


andrewtchin:
check with hagen if this should be cherry picked into 1.3.0

PR #1165, opened 2017-11-28
Remove custom attribute tag (#1161)

jzt:
This change removes the `vic-ova-identifier` custom attribute tag.


jzt:
cherry-picked from `2f67b790`

PR #1222, opened 2017-12-04
Remove Demo VCH Installer Wizard

### LINKED ISSUE #1144: Remove demo VCH deployment from OVA?
With the arrival of the VCH deployment wizard in the vSphere Client, the demo VCH deployment wizard is redundant. @andrewtchin and @morris-jason tell me that no dev work is being done on it in 1.3. I marked it as deprecated in the docs in https://github.com/vmware/vic-product/issues/1106, but @mdubya66 questions whether we should remove it completely. 

@pdaigle, what do you think?  Even if we don't remove it completely, we should probably remove it from the Getting Started page, as users will possibly see the link to the demo installer in that page before they see the full installer in the vSphere Client plugin. It won't be a great user experience if users deploy a limited functionality VCH with the demo installer, only to discover that they could have deployed a full VCH in the client.

From a docs perspective, it will also be a lot tidier if we remove all mention of the demo installer.

### LINKED ISSUE #1214: Delete "Demo VCH deployment Wizard" in Web UI
**VIC Product version:**: `vic-v1.3.0-rc2-2628-9e2a95b6.ova`

According to this [commit](https://github.com/vmware/vic-product/commit/d5f9c2f9a913d093ec2bd1baa06887d419763576) the demo VCH deployment wizard is deprecated, but it still shows up at 2 places in the web UI (on default port 9443) of the RC2 build `vic-v1.3.0-rc2-2628-9e2a95b6` and links to port 1337 of the VIC appliance. The Web UI needs to be adapted accordingly.

image

The demo wizard also does not seem to work in RC2. When I enter the IP address I get the following error message:

image

And using the DNS name of the vCenter Server results in another error message:

image

The credentials work since I can log into the admin portal without issues. On the appliance the directory `/var/log` only contains the following files:

```
ls /var/log/
btmp  journal  lastlog  wtmp
```
The entries in `/etc/resolv.conf` of the VIC appliance are correct and name resolution is working.

andrewtchin:
Fixes #1214 
Fixes #1144


mdubya66:
We are pulling this into 1.3 as the installer doesn't work for the defaults and it is also a call generator. @stuclem let's remove it from the docs

PR #1236, opened 2017-12-07
Fix cert formatting

sidrew:
… when it breaks evenly on 64-character boundary, it injects a newline which borks the cert file.  patch strips blnk lines.


andrewtchin:
I verified this fix. Thank you so much for tracking this down and for your contribution @sidrew!

Issue #1241, opened 2017-12-08
As a doc user, the instructions to install the ui plugin should reference upgrade.sh as an option


mdubya66:
**User Statement:**

As a doc user, the instructions to install the ui plugin should reference upgrade.sh as an option

**Details:**
Followed the install instructions from https://vmware.github.io/vic-product/assets/files/html/1.3/vic_vsphere_admin/plugins_vcsa.html
```
root@docker-client [ ~/vic/ui/VCSA ]# ./install.sh
-------------------------------------------------------------
This script will install vSphere Integrated Containers plugin
for vSphere Client (HTML) and vSphere Web Client (Flex).

Please provide connection information to the vCenter Server.
-------------------------------------------------------------
Enter FQDN or IP to target vCenter Server: vcsa-01a.corp.local
Enter your vCenter Administrator Username: administrator@vsphere.local
Enter your vCenter Administrator Password:

SHA-1 key fingerprint of host 'vcsa-01a.corp.local' is 'F7:68:F0:93:F4:EC:B7:FE:C1:02:3F:F3:AB:62:1A:50:E8:9A:0E:85'
Are you sure you trust the authenticity of this host (yes/no)? yes

-------------------------------------------------------------
Checking existing plugins...
-------------------------------------------------------------
Plugin with key 'com.vmware.vic' is already registered with VC. (Version: 1.3.0.406)
Plugin with key 'com.vmware.vic.ui' is already registered with VC (Version: 1.3.0.406)
-------------------------------------------------------------
Error! At least one plugin is already registered with the target VC.
Please run upgrade.sh instead.
```

**Acceptance Criteria:**
The docs accurately reflect the upgrade workflow

Issue #1240, opened 2017-12-08
as a doc user i need specific information on what file to curl


mdubya66:
**User Statement:**
As a doc user I need to know what file to curl while setting up the UI Plugin.

**Details:**
Figuring out what filename to curl is difficult, for example in https://vmware.github.io/vic-product/assets/files/html/1.3/vic_vsphere_admin/plugins_vcsa.html

**Acceptance Criteria:**
I can copy and paste the curl command and it just works.

stuclem:
@mdubya66 yes, I have gone back and forth on this a few times. We'll just have to make sure that we update the version for every 1.3.x release.

PR #1252, opened 2017-12-12
Added variables to plugin install commands, mentioned upgrade

stuclem:
Fixes https://github.com/vmware/vic-product/issues/1241

Fixes https://github.com/vmware/vic-product/issues/1240

@mdubya66 and @jooskim can you please review? Thanks!


stuclem:
Also fixes https://github.com/vmware/vic-product/issues/1194. @lcastellano can you please review too?

stuclem:
I just tested the procedure in https://github.com/vmware/vic-product/pull/1252/commits/e5bdfa92c77e35fa7a3472d82c9381db2b8a0536 and it worked fine.

stuclem:
@jooskim and @lcastellano should we also include a step in the VCSA plugin install/upgrade topics to clean up the files after you have run the scripts? Afterall, it's not just the plugin files that we copy to the VCSA, but the whole VIC Engine bundle? Is there any reason to leave all that stuff there? For uninstall, we can take the route of unregistering the plug-ins from the VC Managed Object Browser, so there's no need to keep the uninstall script lying around....

stuclem:
I have tested both install and upgrade with the variables. Merging.

Issue #1196, opened 2017-11-30
Run integration tests in parallel


mdharamadas1:
**User Statement:**

As a developer, I want CI build to take optimal time so would like to have integration tests run concurrently.

**Details:**
Need to get away from `pybot` and use something which Mike is trying for VIC repo.

**Acceptance Criteria:**
Integration tests are run in parallel.

PR #1263, opened 2017-12-15
Run integration tests in parallel

mdharamadas1:
Fix for #1196 
Also, added README for tests.


tests/README.md
```diff
@@ -0,0 +1,12 @@
+# Automated Integration Testing
+
+## Approach
+
+Automated tests are designed and developed using [Robot Framework](https://github.com/robotframework/robotframework) and are executed using pybot or pabot.
+These tests are a mix of backend and browser UI tests and are executed in parallel within [Drone](https://github.com/drone/drone) container on CI.
+Browser tests are run using [Docker Selenium Grid](https://github.com/SeleniumHQ/docker-selenium) with headless Firefox nodes which are started per build basis as Drone services.
+
+Common OVA appliance is installed as part of the CI pipeline, before the integration test step, which is shared among all of the tests. It is cleaned up later once the test run is completed.
+All the robot logs and screenshots are published on [Google storage](https://console.cloud.google.com/storage/browser/vic-ci-logs?project=eminent-nation-87317).
```
andrewtchin: Is that bucket for engine too? If so, should these be a separate bucket?

mdharamadas1: Yeah, I'm fine with either buckets. If you feel they should be separate, let me know I will create one.

andrewtchin: As long as there is a link to the correct log bundle in the build output and the log bundle file name shows what repo and build number it is for I guess it's ok in the same

mdharamadas1: Yeah, link is displayed in the logs and looks like below:
`https://console.cloud.google.com/m/cloudstorage/b/vic-ci-logs/o/ova_integration_logs_2885_04bad15bce6f905bcdb556e766133e1d622791ea.zip?authuser=1`

PR #1268, opened 2017-12-18
Add harbor tests

mdharamadas1:
This was a reverted commit, adding back after a fix.


mdharamadas1:
@andrewtchin This was a reverted commit, so the only changes as part of this are the fix for flaky test and make it run concurrently with other tests.

PR #1284, opened 2018-01-05
containers and container VM and infra containers clarification

### LINKED ISSUE #1153: Clarify that Admiral containers are NOT container VMs?
Per comment from @hickeng and @dbarkelew via Slack:

"A question from @dbarkelew today about healthcheck support. He pointed at https://vmware.github.io/vic-product/assets/files/html/1.2/vic_app_dev/provision_containers_portal.html

I'm guessing this is specifically supposed to relate to admiral? however the heading is `container VMs` so it's kind of confusing. We also do not support healthcheck in dockerfile: https://vmware.github.io/vic-product/assets/files/html/1.2/vic_app_dev/dockerfile_support.html

It's not that the instruction cases an error but we do not honor it in the Engine so I think this is misleading."

So, am I to infer from this that the containers that you provision in Admiral/Management Portal are NOT VIC container VMs? Is this the case even if the container host is a VCH? Or, do we need to make a distinction here between containers that you provision in a regular Docker host in Admiral/MP and containers that you provision in a VCH? @sergiosagu and @hickeng can you please clarify? Thanks!

alextopuzov:
Fixes #1153

### DIFF
--- a/docs/user_doc/vic_app_dev/provision_containers_portal.md
+++ b/docs/user_doc/vic_app_dev/provision_containers_portal.md
@@ -1,6 +1,8 @@
 # Provisioning Container VMs in the Management Portal #
 
-You can provision container VMs from the management portal. You can quick-provision containers by using default settings or you can customize your deployment by using the available settings. You can either provision or save as a template your configured container.
+You can provision containers or container VMs from the management portal depending on the target host. If your target host is a VCH, you provision container VMs. If your target host is a Docker host, you provision containers. You can quick-provision by using default settings or you can customize your deployment by using the available settings. You can either provision or save as a template your configured container.
+
+If you provision 
 
 You can provision containers, templates, or images. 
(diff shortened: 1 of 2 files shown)


docs/user_doc/vic_app_dev/provision_containers_portal.md
```diff
@@ -1,6 +1,8 @@
 # Provisioning Container VMs in the Management Portal #
 
-You can provision container VMs from the management portal. You can quick-provision containers by using default settings or you can customize your deployment by using the available settings. You can either provision or save as a template your configured container.
+You can provision containers or container VMs from the management portal depending on the target host. If your target host is a VCH, you provision container VMs. If your target host is a Docker host, you provision containers. You can quick-provision by using default settings or you can customize your deployment by using the available settings. You can either provision or save as a template your configured container.
```
stuclem: "If your target host is a Docker host, you provision containers." Maybe "If your target host is a Docker host, you provision standard containers"?

PR #1289, opened 2018-01-08
Replication Rule Cert fix

### LINKED ISSUE #1154: Update Registry replication cert docs
Per @reasonerjt via Slack, we need to update the registry replication topics: 

----

The only change in Harbor that is visible to VIC is the replication rule:
https://github.com/vmware/harbor/blob/master/docs/user_guide.md#replicating-images
the “Verify remote cert” was moved from system level to target level.

---

alextopuzov:
Fixes #1154


alextopuzov:
Fixed. Check now, pls.

Issue #1270, opened 2017-12-19
Document how move from vApp to RP affects VCH upgrade


stuclem:
Per Slack message from @andrewtchin ages ago:

"(we need) docs for appliance upgrade and vch upgrade with respect to the move to resource pool instead of vapp", "there are some anomalies since a migration procedure wasnt accounted for during the remove RP".

@jzt apparently you're the person who knows about this :)

stuclem:
From @jzt via Slack: 

"I believe the intent was to document that if you are upgrading from a VCH that used a vApp to VCH 1.3 (which uses a resource pool, without the option of a vApp at all), then your old vApp will remain - old containers running in said vApp will continue there, and newly created containers will be deployed into the vApp as well

in the future, upgrading from a vApp VCH to a resource pool VCH will have the behavior described in the 2nd part of that issue (6914) but at the moment, there does not exist a way to migrate containers from one to the other automatically during the upgrade and doing so after the fact is iffy since no resource pool gets created when the VCH gets upgraded"

So, basically, we need to document that upgraded vApp VCHs will remain vApps in 1.3.0, and that migration to resource pools will occur in a future upgrade (1.3.1 or 1.4).

PR #1294, opened 2018-01-09
Mentioned that vApp VCHs remain vApps after upgrade

stuclem:
Fixes https://github.com/vmware/vic-product/issues/1270

@jzt does this cover it? Thanks!


stuclem:
Thanks @jzt !

Issue #949, opened 2017-10-18
Remove workarounds for lost+found issue from docs


stuclem:
Now that https://github.com/vmware/vic/issues/5777 is fixed, the workarounds that @corrieb added to https://github.com/vmware/vic-product/blob/master/docs/user_doc/vic_app_dev/deploy_multiple_docker_compose.md and https://github.com/vmware/vic-product/blob/master/docs/user_doc/vic_app_dev/deploy_single_containervm.md should be removed.

stuclem:
New behviour, high priority. Can be part of general overhaul of the deploying container apps docs.

PR #1302, opened 2018-01-10
Remove haveged examples. Fixes #1301

### LINKED ISSUE #1301: Update docs to remove haveged caveats
This was purportedly fixed in #511, but I found a new place where we need to make an update.

corrieb:
This makes a bunch of updates that should have been made as part of #511. 

In addition, makes some small updates to address #949 for this page.


stuclem:
Fixes https://github.com/vmware/vic-product/issues/1301

PR #1421, opened 2018-02-09
Fix small inconsistency

senk:
Folder `vic` is not in root directory but in the root home folder after following guide.


andrewtchin:
@senk Thank you for your contribution!

Issue #1420, opened 2018-02-08
VIC-Product builds on new drone infra


mdharamadas1:
There are few changes as part of Drone 0.8. This is to investigate those changes and make necessary updates. Also, this will involve making sure all drone secrets are updated and all build steps run as expected.

A/C:
- VIC Product build steps run on new drone infra as expected

andrewtchin:
update and document secrets in vic-internal

PR #1442, opened 2018-02-23
Drone 0.8 updates and changes

mdharamadas1:
Fixes: #1420

This will be merged only after new drone is active.

Outstanding known issues:
- None


.drone.yml
```diff
@@ -14,362 +8,368 @@ pipeline:
...
+  #   image: 'gcr.io/eminent-nation-87317/vic-integration-test:1.42'
+  #   pull: true
+  #   environment:
+  #     BIN: bin
+  #     SHELL: /bin/bash
+  #   commands:
+  #     - 'export COMMIT=${DRONE_COMMIT}'
+  #     - 'echo "Expecting ${DRONE_COMMIT}"'
+  #     - 'git log -5'
+  #     - 'git log -1 --pretty=oneline | grep "^${DRONE_COMMIT}" > /dev/null && echo ''Build matches'' || (git log -1 --pretty=oneline | grep "Merge ${DRONE_COMMIT}" > /dev/null && echo ''Build is of a merge commit'' || (echo ''Build does not match!'' && exit 1))'
```
morris-jason: Is this still needed in drone 8?

mhagen-vmware: @mdharamadas1 we can get rid of the custom clone plugin, but I would like to keep the display status for the time being until we trust the clone step more

mdharamadas1: done

PR #1458, opened 2018-03-06
Added note about network folders to ops user config doc

stuclem:
Doc update resulting from Bugzilla issue 2064396. Added clarification to the entry about datastore folders. We already state in the table that the datacenter role should not propagate.

@hickeng @dbarkelew @lcastellano does this cover it? Thanks!

### DIFF
--- a/docs/user_doc/vic_vsphere_admin/ops_user_manual.md
+++ b/docs/user_doc/vic_vsphere_admin/ops_user_manual.md
@@ -88,7 +88,7 @@ The following table lists which roles to assign to which type of inventory objec
 <tr>
 <th>Inventory Object</th>
 <th>Role to Assign</th>
-<th>Propagate</th>
+<th>Propagate?</th>
 </tr>
 </thead>
 <tbody>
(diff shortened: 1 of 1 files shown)


hickeng:
@stuclem We need either:
1. the datacenter role to propagate, or
2. the DVS in a network folder to which we can assign the Read-Only/propagate role.

What we're trying to accomplish is having Read-Only propagate down the network folders - either of those approaches will do so as the assignment of _any_ permission other than read-only seems to include read-only.

Additionally we need the following privileges added - are these in a different PR?:
1. Add Virtual Machine->Configuration->Add Existing Disk to the Datacenter role
2. Add Global->Enable/Disable Method to the Endpoint role

stuclem:
**

stuclem:
See https://github.com/vmware/vic/issues/7481 for some additional information.

stuclem:
Thanks @hickeng. I attempted to address this in https://github.com/vmware/vic-product/pull/1458/commits/ad902ee333aa97534cec1d887a21fb6c0b2e4c65. Does that cover it?

stuclem:
Thanks @hickeng. Added @lgayatri as a reviewer.

PR #1471, opened 2018-03-09
Add timestamps for tdnf install steps

andrewtchin:


morris-jason:
This is ugly, but the current build log isn't exactly pretty. LGTM, will be replaced by the upgrade improvement pr.

Issue #1443, opened 2018-02-26
Troubleshooting doc for non-appearing UI issue


stuclem:
Per https://github.com/vmware/vic-ui/issues/299#issuecomment-368060252 we need to include this issue as a TS topic in the core docs.

PR #1484, opened 2018-03-19
Added link to UI bug KB

stuclem:
Fixes https://github.com/vmware/vic-product/issues/1443. Since we already have a troubleshooting topic for non-appearing UI problems, I integrated this info in there rather than creating a new topic. 

@jooskim does this do it?


stuclem:
Thanks @jooskim

Issue #1360, opened 2018-01-30
vic-product: multiple datacenter


mhagen-vmware:
https://github.com/vmware/vic/blob/master/tests/manual-test-cases/Group5-Functional-Tests/5-10-Multiple-Datacenter.robot

mdharamadas1:
Added test in `vic-ui` repo

PR #1486, opened 2018-03-19
Add nightly test for multiple datacenter

mdharamadas1:
Fix for #1360

Verified this test passes on my local.

### DIFF
--- a/tests/resources/OVA-Util.robot
+++ b/tests/resources/OVA-Util.robot
@@ -33,8 +33,7 @@ Set Test OVA IP If Available
     Run Keyword If  ${rc} == 0  Set Environment Variable  OVA_IP  ${output}
     [Return]  ${rc}
 
-Install VIC Product OVA
-    [Tags]  secret
+Install VIC Product OVA Only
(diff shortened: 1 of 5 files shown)


tests/resources/OVA-Util.robot
```diff
@@ -49,9 +49,17 @@ Install VIC Product OVA
     \   ${ova-ip}=  Run Keyword If  ${status}  Set Variable  ${ip}  ELSE  Set Variable  ${ova-ip}
 
     Wait For Register Page  ${ova-ip}
+    Set Environment Variable  OVA_IP  ${ova-ip}
+
+    [Return]  ${ova-ip}
+
+Install VIC Product OVA
```
andrewtchin: Should this be renamed to `Install VIC Product OVA and Initialize`?

mdharamadas1: Yeah I would prefer that too but it will break the test API and don't wanted a lot of changes. Anyway, we are adding these tests in `vic-ui` repo instead, so closing this one.

mdharamadas1:
Instead adding these tests in `vic-ui` repo.

PR #1495, opened 2018-03-20
WIP Add harbor and admiral to an internal docker network

### LINKED ISSUE #419: Harbor and Admiral services should communicate via internal network on .ova
Harbor needs to call PMS API hosted on Admiral endpoint, and Admiral UI backend will wrap Harbor's API for UI to render registry related components.

Currently they communicate via the external IP, which works so far, but ideally they should communicate via an internal network.

In bootstrap it maybe possible to call "docker network create/connect" to address this issue.

morris-jason:
Fixes #419

Containers are successfully added to a internal network, but there are unintended consequences of changing harbor.tab.url to accomplish this. 
Waiting for responses from @reasonerjt and @lazarin on #419.

### DIFF
--- a/installer/build/ova-manifest.json
+++ b/installer/build/ova-manifest.json
@@ -92,6 +92,11 @@
   },
   {
     "type": "file",
+    "source": "scripts/harbor/configure_internal_network.sh",
+    "destination": "/etc/vmware/harbor/configure_internal_network.sh"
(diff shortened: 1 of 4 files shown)


morris-jason:
Closing this in favor of a cert-pinning approach for mitigating mitm.

PR #1511, opened 2018-03-26
Don't always collect ovfenv in support logs

andrewtchin:


dbarkelew:
Can this be changed so by default the information that isn't sensitive is captured? Like the IP Address and ports for the services in the off changes the deployment is not using the default ports. 

This might be a separate task to track. Having a default validation check for the certificate format to make sure the admiral service has the certificate in the required format.

Issue #1422, opened 2018-02-09
Add time synchronization docs


andrewtchin:
Currently the appliance receives NTP configuration from DHCP. If using static IP, NTP might not be configured and the user may experience clock skew. The effects of clock skew could be certificate errors and errors obtaining PSC tokens resulting in failure of Admiral. 

Users should enable guest time synchronization through vSphere Client or set NTP manually by SSHing to the appliance and running `timedatectl set-ntp true`. NTP servers can be customized in `/etc/systemd/timesyncd.conf`.  

Patrick reported that his environment required a Shutdown Guest OS and Power On as opposed to a Reboot to apply the guest time synchronization setting from vSphere Client.

https://kb.vmware.com/s/article/1014038
![image](https://user-images.githubusercontent.com/1649165/36036203-ded6a750-0d7e-11e8-9c56-2d9d24dca8d5.png)

Related https://github.com/vmware/vic-product/issues/1390

PR #1513, opened 2018-03-27
Added info about configuring NTP in appliance

stuclem:
Fixes https://github.com/vmware/vic-product/issues/1422.

@andrewtchin does this cover it? Since we already had a TS topic on clock skews in the VCH, I added the info there.


stuclem:
Thanks @andrewtchin

PR #1524, opened 2018-03-29
Adding info about NFS UID/GID and root user

stuclem:
Fixes https://github.com/vmware/vic-product/issues/1464 (possibly, but probably needs reworking).

@matthewavery and @dbarkelew I really am not very sure about this. This first attempt is just that. Hopefully we can flesh it out in this review.


docs/user_doc/vic_app_dev/using_volumes_with_vic.md
```diff
@@ -163,6 +163,8 @@ If your volume store is in an NFS share point, sharing volumes between container
 
 <pre>docker volume create --opt volumestore=<i>nfs_volumestore_name</i></pre>
 
+**NOTE**: vSphere Integrated Containers mounts NFS volumes as `root`. Consequently, if containers are to run as non-root users, the volume  store must be configured with the correct permissions so that the non-root users can access it.
```
matthewavery: In this case the permissions are going to come from the nfs server. So if a container run's as non-root and is trying to write into the nfs mount then the permissions from the server side of things must allow the accessing user the ability to perform the requested actions. i.e. the `other` (also known as world) permissions must be set, or the user accessing the NFS mount should be part of the same `Group` that owns the share(I think this ends up being root for us as well). Something that helps with this is configuring your share point with an anonymous user. Another thing that can help alot potentially, `root_squash` is designed to map the root user to the anonymous user. `all squash` will map all uid/gids to the anonymous uid/gid.

matthewavery: for reference from my local exports man page:

```
       root_squash
              Map requests from uid/gid 0 to the anonymous uid/gid. Note that this does not apply to any other uids or gids that might be equally sensitive, such as user bin or group staff.

       no_root_squash
              Turn off root squashing. This option is mainly useful for diskless clients.

       all_squash
              Map all uids and gids to the anonymous user. Useful for NFS-exported public FTP directories, news spool directories, etc. The opposite option is no_all_squash, which is the default setting.

       anonuid and anongid
              These  options explicitly set the uid and gid of the anonymous account.  This option is primarily useful for PC/NFS clients, where you might want all requests appear to be from one user. As an example, consider the export entry for /home/joe in the
              example section below, which maps all requests to uid 150 (which is supposedly that of user joe).
```

matthewavery: So in this case the volume store can be configured with a uid/gid for creation and reading at the portlayer level. But the configuration of the sharepoint will still need to be customer specific and setup specfic. It will likely need to map the root user to the anon since we will have many containers potentially attempting to read(or write) to the same location. This kind of configuration would not be done at `vic-machine` time. it must be done on the nfs server

docs/user_doc/vic_vsphere_admin/volume_stores.md
```diff
@@ -129,10 +129,13 @@ To specify an NFS share point as a volume store, use the `nfs://` prefix and the
 <a id="nfsoptions"></a>
 You can also specify the URL, UID, and GID of a shared NFS mount point when you specify an NFS share point. Connections are made over TCP. If you do not specify a UID and GID, vSphere Integrated Containers Engine uses the `anon` UID and GID when creating and interacting with the volume store. The `anon` UID and GID is 1000:1000.
 
-**NOTE**: If your NFS server uses a different `anon` UID/GID to the default, you must specify the UID/GID in the `--volume-store` option. Configuring a VCH to use a different default `anon` UID/GID for NFS volume stores is not supported.
-
 <pre>--volume-store nfs://<i>datastore_address</i>/<i>path_to_share_point</i>?uid=1234&gid=5678&:<i>nfs_volume_store_label</i></pre> 
 
+**NOTES**: 
+
+- If your NFS server uses a different `anon` UID/GID to the default, you must specify the UID/GID in the `--volume-store` option. Configuring a VCH to use a different default `anon` UID/GID for NFS volume stores is not supported.
```
matthewavery: This is to facilitate the creation and reading of the volume store at the portlayer. If we cannot read or write to it during portlayer time then we will not be able to use/make the volume store. For containers the user that is running the process in the container will still need to have the correct permissions on the mount to read and write. See above comment about UID/GID squashing and unix permissions.

docs/user_doc/vic_vsphere_admin/volume_stores.md
```diff
@@ -129,10 +129,13 @@ To specify an NFS share point as a volume store, use the `nfs://` prefix and the
...
+- vSphere Integrated Containers mounts NFS volumes as `root`. Consequently, if you specify a UID/GID, it must be valid for `root`. Additionally, if containers are to run as non-root users, the export of the volume must grant the correct permissions to the non-root users so that they can access the volume store.
```
matthewavery: The `tether` will mount it as root. but the portlayer may request the mount with the specified UID/GID. Per the portions starting with `the export of the volume must grant the correct...` the information above about sqaushing UID/GIDs as well as unix permissions applies here. There are several ways to assign "correct" permissions to the share point for VIC, but this will depend on the customer, their use case, and their policy for configuring NFS.

docs/user_doc/vic_app_dev/using_volumes_with_vic.md
dbarkelew: Was able to configure "no_root_squash" for the NFS to work on a NetApp. Note: the root user also must be a member of the GID the user configures the VCH to use.

docs/user_doc/vic_vsphere_admin/volume_stores.md
```diff
@@ -129,10 +129,13 @@ To specify an NFS share point as a volume store, use the `nfs://` prefix and the
 <a id="nfsoptions"></a>
 You can also specify the URL, UID, and GID of a shared NFS mount point when you specify an NFS share point. Connections are made over TCP. If you do not specify a UID and GID, vSphere Integrated Containers Engine uses the `anon` UID and GID when creating and interacting with the volume store. The `anon` UID and GID is 1000:1000.
 
-**NOTE**: If your NFS server uses a different `anon` UID/GID to the default, you must specify the UID/GID in the `--volume-store` option. Configuring a VCH to use a different default `anon` UID/GID for NFS volume stores is not supported.
-
 <pre>--volume-store nfs://<i>datastore_address</i>/<i>path_to_share_point</i>?uid=1234&gid=5678&:<i>nfs_volume_store_label</i></pre>
```
dbarkelew: Maybe I am wrong but I think there is an extra & in the example. I would expect this to look something like --volume-store nfs://<i>datastore_address</i>/<i>path_to_share_point</i>?uid=1234&gid=5678:<i>nfs_volume_store_label</i>

docs/user_doc/vic_vsphere_admin/volume_stores.md
dbarkelew: I am not following this note why does anon come into play? I was under the impression that the VCH and containers should be using users that are part of the GID and there shouldn't be a need for anon permissions or changes to anon.

docs/user_doc/vic_vsphere_admin/volume_stores.md
stuclem: Fixed in https://github.com/vmware/vic-product/pull/1524/commits/4ca4b1b70a81d642485a9f272ca844fd55bab18f.

docs/user_doc/vic_vsphere_admin/volume_stores.md
stuclem: Attempted to recreate all this in https://github.com/vmware/vic-product/pull/1524/commits/4ca4b1b70a81d642485a9f272ca844fd55bab18f.

docs/user_doc/vic_vsphere_admin/volume_stores.md
stuclem: Attempted to recreate all this in https://github.com/vmware/vic-product/pull/1524/commits/4ca4b1b70a81d642485a9f272ca844fd55bab18f.

docs/user_doc/vic_app_dev/using_volumes_with_vic.md
stuclem: Attempted to recreate all this in https://github.com/vmware/vic-product/pull/1524/commits/4ca4b1b70a81d642485a9f272ca844fd55bab18f.

stuclem:
https://github.com/vmware/vic-product/pull/1524/commits/4fc9a33c6e2bdf70568e0478993a50d4d7f80764 fixes https://github.com/vmware/vic-product/issues/1641.

docs/user_doc/vic_vsphere_admin/volume_stores.md
```diff
@@ -51,6 +51,16 @@ To use shared NFS volume stores, it is recommended that the NFS share points tha
...
+
+Permissions are determined by the NFS server. If a container runs as non-root and it attempts to write into the NFS mount, the permissions on the server side must grant the accessing user the ability to perform the requested actions. To allow this, the `other` (also known as `world`) permissions must be set. Alternatively, the user that is accessing the NFS mount should be part of the same `Group` that owns the share. To facilitate this, you can configure the share point with an anonymous user. You can also use `root_squash`, which is designed to map the root user to the anonymous user. Using `all squash` maps all UIDs/GIDs to the anonymous UID/GID.
+
+In this case, you can configure the volume store with a UID/GID for creation and reading. The configuration of the sharepoint is dependent on your setup, but it will most likely need to map the `root` user to the `anon` user because there will be many containers potentially attempting to read or write to the same location. You make this configuration on the NFS server.
```
matthewavery: This is a guaranteed must. Since the containers will only attempt a mount with the UID of 0(root) then the nfs server provided will need to have the sharepoint(nfs target) configured as an anonymous target with atleast root level squashing enabled. I think that is somewhat clear here, but I do not want to leave room for misinterpretation at this point. This could change eventually when we allow nfs client mount args with the create command(if we get to that soon), but for now squashing is a must afaik.

docs/user_doc/vic_app_dev/container_operations.md
```diff
@@ -41,7 +41,7 @@
 |`attach`|[Attach to a container](https://docs.docker.com/engine/reference/commandline/attach/)|Yes, since 1.0|
 |`container list`|[List Containers](https://docs.docker.com/engine/reference/api/docker_remote_api_v1.22/#list-containers)|Yes, since 1.0|
 |`container resize`|[Resize a container](https://docs.docker.com/engine/reference/api/docker_remote_api_v1.23/#resize-a-container-tty)|Yes, since 1.0|
-|`cp`|[Copy files or folders between a container and the local filesystem](https://docs.docker.com/engine/reference/commandline/cp/)|Yes, since 1.2. You cannot copy to or from an NFS volume. You cannot copy from an unstarted container.|
+|`cp`|[Copy files or folders between a container and the local filesystem](https://docs.docker.com/engine/reference/commandline/cp/)|Yes, since 1.2. You cannot copy to an NFS volume that is not in use by a running container. You cannot copy from an unstarted container.|
```
matthewavery: `You cannot copy from an unstarted container.` is not accurate. This is only true for NFS volumes. the container must be running for NFS. Otherwise copying from a volume or a container when it is not powered on should work perfectly well. The only special case is NFS and the container powerstate(needs to be on).

docs/user_doc/vic_app_dev/using_volumes_with_vic.md
```diff
@@ -163,6 +165,8 @@ If your volume store is in an NFS share point, sharing volumes between container
 
 <pre>docker volume create --opt volumestore=<i>nfs_volumestore_name</i></pre>
 
+**NOTE**: vSphere Integrated Containers mounts NFS volumes as `root`. Consequently, if containers are to run as non-root users, the volume  store must be configured with the correct permissions so that the non-root users can access it. For information about how to configure NFS volume stores for non-root users, see [About NFS Volume Stores and Permissions](../vic_vsphere_admin/volume_stores.md#nfs_perms) in *Install, Deploy, and Maintain the vSphere Integrated Containers Infrastructure*.
```
matthewavery: This definitely looks great, Not sure if we should mention the need for squashing root here. Since it links to the share discussion probably not, but the need for squashing is required tmk.

stuclem: Added it anyway in https://github.com/vmware/vic-product/pull/1524/commits/dd458a61ba96ef516a3e5db6a110b78edadceb58.

docs/user_doc/vic_vsphere_admin/volume_stores.md
stuclem: Updated in https://github.com/vmware/vic-product/pull/1524/commits/dd458a61ba96ef516a3e5db6a110b78edadceb58.

docs/user_doc/vic_app_dev/container_operations.md
stuclem: Updated in https://github.com/vmware/vic-product/pull/1524/commits/dd458a61ba96ef516a3e5db6a110b78edadceb58.

docs/user_doc/vic_app_dev/using_volumes_with_vic.md
```diff
@@ -163,6 +165,8 @@ If your volume store is in an NFS share point, sharing volumes between container
 
 <pre>docker volume create --opt volumestore=<i>nfs_volumestore_name</i></pre>
 
+**NOTE**: vSphere Integrated Containers mounts NFS volumes as `root`. Consequently, if containers are to run as non-root users, the volume  store must be configured with the correct permissions so that the non-root users can access it, with at least root level squashing enabled. For information about how to configure NFS volume stores for non-root users, see [About NFS Volume Stores and Permissions](../vic_vsphere_admin/volume_stores.md#nfs_perms) in *Install, Deploy, and Maintain the vSphere Integrated Containers Infrastructure*.
```
dbarkelew: Shouldn't it be root level squashing disabled?

docs/user_doc/vic_vsphere_admin/volume_stores.md
```diff
@@ -51,6 +51,16 @@ To use shared NFS volume stores, it is recommended that the NFS share points tha
...
+
+Permissions are determined by the NFS server. If a container runs as non-root and it attempts to write into the NFS mount, the permissions on the server side must grant the accessing user the ability to perform the requested actions. To allow this, the `other` (also known as `world`) permissions must be set. Alternatively, the user that is accessing the NFS mount should be part of the same `Group` that owns the share. To facilitate this, you can configure the share point with an anonymous user. You can also use `root_squash`, which is designed to map the root user to the anonymous user. Using `all squash` maps all UIDs/GIDs to the anonymous UID/GID.
+
+In this case, you must configure the volume store with a UID/GID for creation and reading. The configuration of the sharepoint is dependent on your setup, but it will need to map the `root` user to the `anon` user with at least root level squashing enabled. This is because there will be many containers potentially attempting to read or write to the same location. You make this configuration on the NFS server.
```
dbarkelew: Pretty sure this needs to be root level squashing disabled. No not want root mapping to anon user.

dbarkelew:
There are a few NFS configurations that should work. Copying my suggestion from an email, documentation would cover these better if said something like "If the NFS is set to no_squash_root then root user/group needs permissions to the NFS share/export. If squash_root is enabled, then the anon user/group must have permissions to the NFS share/export”.

PR #1610, opened 2018-04-17
Don't download the OVA each time in nightly

andrewtchin:


mdharamadas1:
@andrewtchin Can you add below line in `OVA Setup` and merge this?
`Set Environment Variable  DRONE_BUILD_NUMBER  0`

PR #1630, opened 2018-04-20
Change owner of CA certificate and token file after they are copied to Harbor

reasonerjt:
After ca.crt and tokens.properties are copied to Harbor's directory the owner changed to root.

This PR changes the owner to 10000:10000 after it's copied.

This fixes the build issue in master branch.


reasonerjt:
The error in test case `Verify default harbor registry is displayed` has been fixed, but there still a failure in tc `Push an image to harbor and create a container`.  

The reason for second failure is that per PM's request Harbor will show cardview of images by default, and there's another issue opened to track this problem:
https://github.com/vmware/vic-product/issues/1632

reasonerjt:
Merging this PR so @ninjadq can check the UI TC failure.

PR #1631, opened 2018-04-20
Fix Nightly DRS test

mdharamadas1:
Fixed incorrect Robot keyword


mdharamadas1:
@singhshwetaQE Thanks for pointed that out, corrected now.

PR #1661, opened 2018-04-27
Corrected name of vic-machine-server.service

stuclem:
While checking the services running in the RC2 OVA, I noticed that the doc misnames the `vic-machine-server.service`.

@zjs has this name always been wrong in the docs, or did it change in 1.4? Thanks!


andrewtchin:
I think I renamed it in this release. Current version is indeed using `-`

PR #1684, opened 2018-05-02
unify harbor db and clair db password in first boot

wy65701436:
As for Harbor DB migrates to pgsql, the clair DB needs to share the same password with harbor DB. This commit is to fix it in vic env.

### DIFF
--- a/installer/build/scripts/harbor/configure_harbor.sh
+++ b/installer/build/scripts/harbor/configure_harbor.sh
@@ -138,9 +138,10 @@ configureHarborCfg ssl_cert $cert
 configureHarborCfg ssl_cert_key $key
 configureHarborCfg secretkey_path $data_dir
 
-# Set MySQL and Clair DB passwords on first boot
-configureHarborCfgOnce db_password "$(genPass)"
-configureHarborCfgOnce clair_db_password "$(genPass)"
+# Set Harbor DB and Clair DB passwords on first boot
+random_pwd=$(genPass)
+configureHarborCfgOnce db_password "$random_pwd"
+configureHarborCfgOnce clair_db_password "$random_pwd"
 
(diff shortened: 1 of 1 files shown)


installer/build/scripts/harbor/configure_harbor.sh
```diff
@@ -138,9 +138,10 @@ configureHarborCfg ssl_cert $cert
 configureHarborCfg ssl_cert_key $key
 configureHarborCfg secretkey_path $data_dir
 
-# Set MySQL and Clair DB passwords on first boot
-configureHarborCfgOnce db_password "$(genPass)"
-configureHarborCfgOnce clair_db_password "$(genPass)"
+# Set Harbor DB and Clair DB passwords on first boot
```
andrewtchin: What happens with an upgraded config? `configureHarborCfgOnce` would not overwrite an existing managed config line so if after migration the `db_password` or `clair_db_password` were tagged with the managed by line, lines 143-144 would have no effect.

andrewtchin: Looks like these lines are tagged after upgrade https://github.com/vmware/vic-product/blob/master/installer/build/scripts/upgrade/upgrade-harbor.sh#L126

andrewtchin: Is this behavior acceptable?

wy65701436: @andrewtchin could I open anther PR to fix the upgrade issue later? It's because that we are still working on the implementation of DB migration scrip. It may take at least one sprint as the big change on Harbor DB in the next release.

wy65701436: This PR could fix the harbor launch issue, and do not block vic CI.

Issue #1655, opened 2018-04-27
Document that you still cannot deploy OVA from H5 client in 6.7, even though it appears to work


stuclem:
I just tested this with RC2 and it succeeded. We need to check whether other tasks that we currently document as only being possible in the Flex client are possible in H5 6.7.

stuclem:
For an example of what happens if you use the h5 client, see https://github.com/vmware/vic-product/issues/1660.

stuclem:
What we need to make clear is that this is a problem with the H5 client, not with VIC! @andrewtchin do you agree?

PR #1689, opened 2018-05-03
Added TS topic about deploying OVA using 6.7 H5 client

stuclem:
Fixes https://github.com/vmware/vic-product/issues/1655


andrewtchin:
ovftool also works

PR #1693, opened 2018-05-03
Pull release version of vic-machine-server

morris-jason:
VIC Appliance Checklist:
- [ ] Added tests
- [ ] Tests passing
- [ ] Updated documentation

Fixes #397. Relies on vmware/vic#7852.

### DIFF
--- a/installer/build/build-ova.sh
+++ b/installer/build/build-ova.sh
@@ -82,7 +82,7 @@ url=$(gsutil ls -l "gs://vic-engine-builds" | grep -v TOTAL | grep vic_ | sort -
 setenv VICENGINE "$url"
 
 #set Harbor
-url=$(gsutil ls -l "gs://harbor-builds" | grep -v TOTAL | grep offline-installer | sort -k2 -r | (trap '' PIPE; head -n1) | xargs | cut -d ' ' -f 3 | sed 's/gs:\/\//https:\/\/storage.googleapis.com\//')
+url=$(gsutil ls -l "gs://harbor-builds" | grep -v TOTAL | grep offline-installer | grep -v offline-installer-latest | sort -k2 -r | (trap '' PIPE; head -n1) | xargs | cut -d ' ' -f 3 | sed 's/gs:\/\//https:\/\/storage.googleapis.com\//')
 setenv HARBOR "$url"
 
 export BUILD_DCHPHOTON_VERSION="1.13"
--- a/installer/scripts/ci-build.sh
+++ b/installer/scripts/ci-build.sh
@@ -47,7 +47,21 @@ fi
 if [ -n "${VIC_MACHINE_SERVER}" ]; then
   OPTIONS="$OPTIONS --vicmachineserver $VIC_MACHINE_SERVER"
 elif [[ "$DRONE_BRANCH" == *"releases/"* ]]; then
-  vicmachineserver_release="latest"
+
+  # Listing container tags requires permissions
+  if [ -z "$(gcloud auth list --filter=status:ACTIVE --format='value(account)')" ]; then
+    if [ -z "${GOOGLE_KEY}" ]; then
+      echo "No google service account key found..."
+      exit 1
+    fi
+    echo "Attempting to login with google account service key"
+    KEY_FILE=".tmp.token"
+    cat "${GOOGLE_KEY}" > ${KEY_FILE}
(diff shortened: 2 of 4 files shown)


installer/scripts/ci-build.sh
```diff
@@ -47,7 +47,21 @@ fi
...
+  if [ -z "$(gcloud auth list --filter=status:ACTIVE --format='value(account)')" ]; then
+    if [ -z "${GOOGLE_KEY}" ]; then
+      echo "No google service account key found..."
+      exit 1
+    fi
+    echo "Attempting to login with google account service key"
+    KEY_FILE=".tmp.token"
+    echo "${GOOGLE_KEY}" > ${KEY_FILE}
+    gcloud auth activate-service-account --key-file ${KEY_FILE} || (echo "Login with service account key failed..." && exit 1)
+    rm ${KEY_FILE}
```
andrewtchin: maybe `-f`

installer/build/build-ova.sh
```diff
@@ -82,7 +82,7 @@ url=$(gsutil ls -l "gs://vic-engine-builds" | grep -v TOTAL | grep vic_ | sort -
 setenv VICENGINE "$url"
 
 #set Harbor
-url=$(gsutil ls -l "gs://harbor-builds" | grep -v TOTAL | grep offline-installer | sort -k2 -r | (trap '' PIPE; head -n1) | xargs | cut -d ' ' -f 3 | sed 's/gs:\/\//https:\/\/storage.googleapis.com\//')
+url=$(gsutil ls -l "gs://harbor-builds" | grep -v TOTAL | grep offline-installer | grep -v offline-installer-latest | sort -k2 -r | (trap '' PIPE; head -n1) | xargs | cut -d ' ' -f 3 | sed 's/gs:\/\//https:\/\/storage.googleapis.com\//')
```
zjs: Should the other line like this be updated too?

https://github.com/vmware/vic-product/blob/2d0f7063ca8ddf024ef2002a955b4a26a59c5d19/installer/scripts/ci-build.sh#L36

installer/scripts/ci-build.sh
```diff
@@ -47,7 +47,21 @@ fi
...
+  fi
+
+  vicmachineserver_release="$(gcloud container images list-tags gcr.io/eminent-nation-87317/vic-machine-server --filter='tags~.' | grep -v DIGEST | awk '{print $2}' | sed -rn '/^v([0-9]\.?){3}(-rc[0-9])?$/p' | head -n 1)"
```
zjs: 1. It seems like this will get the latest tag, but not necessarily the one with the highest version number. Is that what we want? (Our whole release process seems to get really messy if we consider cases like making a 1.4.1 to fix a security issue after 1.5.0 ships. Maybe that's not something we want to worry about at this point.)

2. Is `v([0-9]\.?){3}(-rc[0-9])?` an expression we're using other places? If so, we can use it for now. If not, it seems a bit fragile (e.g., it'll break for `v1.10.0` or `v1.4.0-rc10`). Also, it matches some things that it probably shouldn't (e.g., `v123`, `v1.4.0.`, `v1.10`).

installer/build/build-ova.sh
morris-jason: `harbor-offline-installer-lastest` artifacts __should__ only be posted to harbor-builds, not harbor releases.

installer/scripts/ci-build.sh
morris-jason: 1) I had this conversation earlier with Andrew. I don't see a way to automate the pulling in of specific version numbers:
  - We can't pull in exact version numbers, because harbor is a version ahead of us:
  - We can't say pull in exact numbers or higher, because in 1.3.1 we wanted harbor 1.3.0 even though harbor 1.4.0 exists.
In general, I don't know if we can do much here, but I'm open to conversations.
I don't know if automatically pulling in release tags is even the best idea. A better solution is to find an easier way to specify the proper component numbers outside of drone, possible in tag metadata.

2) Yeah.... I overlooked that regex. It is not used in other places. I can make it more resilient if we want.

zjs: > A better solution is to find an easier way to specify the proper component numbers outside of drone, possible in tag metadata.

Something along these lines would be good to address eventually, but this sounds like a separate project though.

morris-jason: Yes, exactly. Gandalf could also be helpful here ;)

Issue #1709, opened 2018-05-07
VIC upgrade scenario breaks by Harbor DB changes(master)


wy65701436:
As for harbor has migrated DB to pgsql, the migrator won't work for the master build, and it breaks the VIC upgrade test in nightly suite.

Harbor team will provide an new DB migrator, let's use this issue to track.

mdharamadas1:
FYI, this is currently a blocker for night test suite as its impacting upgrade tests.

sgairo:
@mdharamadas1 Do you feel comfortable giving it an estimate?

mdharamadas1:
@sgairo I can't, probably @wy65701436 ?

mdubya66:
making this a p0 since it's blocking successful nightlies.

wy65701436:
@andrewtchin @morrisjason, please update the builder script to pick up harbor stable build from https://storage.googleapis.com/harbor-builds/master.stable, it holds the URL of gcr. 

Harbor team could make sure the build in the master.stable passed our full CI, and won't break the upgrade testing in nightly.

morris-jason:
@reasonerjt wy65701436 I do not believe this is the correct approach.

This file links to `https://storage.googleapis.com/harbor-releases/release-1.5.0/harbor-offline-installer-v1.5.0.tgz`, but we expect it to link to the latest stable **master** build, ie `https://storage.googleapis.com/harbor-builds/harbor-offline-installer-v1.5.0-build.1234.tgz`

reasonerjt:
@morris-jason,  thanks for pointing it out.

Currently the `master.stable` points to a 1.5 branch build because the upgrade master build, and we expect it points to build on master branch once we verified the upgrade.

hypothetically, if you wanna ship a 1.6 release tomorrow, you can package this 1.5 build of Harbor and nothing will break.

The idea is that we explicitly provide a Harbor build to tell you that **this is the build should be integrated**.

Are you suggesting the filename `master.stable` is misleading?  We may change it to `vic.master.stable` to indicate it's the build to be integrated with vic-product `master` branch, but it's a little trivial.

morris-jason:
@reasonerjt my concerns that the stable file should contain a link to the harbor-builds bucket, not harbor-releases, is trivial. 

The overall concern here is that, because this is now a very manual process, the ova testing will occur on old versions of harbor if the stable file is not updated in a timely manner.

[...]

Issue #1568, opened 2018-04-05
Shared ca_download directory


andrewtchin:
**User Statement:**
As a developer I want to simplify the startup process of the appliance.

**Details:**
https://sourcegraph.com/github.com/vmware/vic-product/-/blob/installer/build/scripts/harbor/configure_harbor.sh#L103
https://sourcegraph.com/github.com/vmware/vic-product/-/blob/installer/build/scripts/admiral/configure_admiral.sh#L86
https://sourcegraph.com/github.com/vmware/vic-product/-/blob/installer/build/scripts/fileserver/configure_fileserver.sh#L23

**Acceptance Criteria:**
- [ ] create a single ca_download_dir
- [ ] tell admiral and harbor where to reference it as a read only volume
- [ ] update DESIGN.md

PR #1719, opened 2018-05-08
Add integration document for harbor

wy65701436:


installer/docs/integration/harbor.md
```diff
@@ -0,0 +1,128 @@
...
+  `vmware/clair-photon`
+  `vmware/notary-server-photon`
+  `vmware/notary-signer-photon`
+  `vmware/nginx-photon`
+  `vmware/harbor-log`
+  `vmware/harbor-jobservice`
+  `vmware/harbor-ui`
+  `vmware/mariadb-photon`
+  `vmware/postgresql-photon`
+
```
morris-jason: @wy65701436 I think there are a few more containers missing here:
```
vmware/harbor-adminserver
vmware/harbor-db
vmware/harbor-migrator
vmware/registry-photon
```

installer/docs/integration/harbor.md
```diff
@@ -0,0 +1,128 @@
...
+    - Release versions MUST be retained. Where will previously released versions be stored?
+      `gcr.io/eminent-nation-87317/harbor-releases/${release_branch}/${release_build}`
+      
+- What is the component's naming scheme for development and release builds?
+  Component is delivered as a tgz file named `harbor-offline-installer-build.${drone_job_id}.tgz` for development and
```
morris-jason: Can you make a note here about `latest.build` and `harbor-offline-installer-latest` in the dev bucket?

installer/docs/integration/harbor.md
```diff
@@ -0,0 +1,128 @@
+# VIC Appliance Component Integration Documentation
+
+## VIC Machine Server
```
andrewtchin: `Harbor`

installer/docs/integration/harbor.md
```diff
@@ -0,0 +1,128 @@
...
+  `/storage/db`, `/storage/data` and `/storage/log`
+
+    - What is the default size in GB?
+    N/A
+
+- Describe the requirements the platform must provide for the component
+    - Examples: OS packages, certificates, token files, configuration
+    - Specify the location the component expects any dependencies to be on the filesystem:
+  Requires values of hostname, ui_url_protocol, ssl_cert, ssl_cert_key, secretkey_path and admiral_url in `/storage/data/harbor/harbor.cfg`
+  Requires `ca.crt` in `/storage/data/harbor/ca_download` with permission 10000:10000
```
andrewtchin: May change in the future pending https://github.com/vmware/vic-product/issues/1568

installer/docs/integration/harbor.md
```diff
@@ -0,0 +1,128 @@
...
+
+- Specify networking requirements
+    - External networking:
+      Port 443/9443 TCP allowed
+    - Will the component connect to the appliance's `vic-appliance` network?
+      https://docs.docker.com/compose/networking/#use-a-pre-existing-network
+      NO
+
+- Does the component require any other components to be running before it starts?
+  NO
```
andrewtchin: What about PSC? Is Admiral required to be started?

installer/docs/integration/harbor.md
```diff
@@ -0,0 +1,128 @@
...
+
+- Are there any steps that need to be performed before starting the component?
+  NO
```
andrewtchin: What about PSC?

installer/docs/integration/harbor.md
```diff
@@ -0,0 +1,128 @@
...
+
+- Should a script be run before starting the component?
+  Ensure the `/etc/vmware/harbor/prepare` script is executed.
```
andrewtchin: Is it OK to run this every time?

installer/docs/integration/harbor.md
```diff
@@ -0,0 +1,128 @@
...
+  Built by `vmware/harbor` https://github.com/vmware/harbor/blob/master/.drone.yml
+
+- Where will the VIC appliance build obtain component artifacts?
+  Load from harbor offline installer
+  
+    - Link to bucket that will contain development builds: `gcr.io/eminent-nation-87317/harbor-builds`
+    - Link to bucket that will contain tagged builds for releases (including release candidates):
+      `gcr.io/eminent-nation-87317/harbor-releases/${branch_name}`
+    - How will the VIC appliance build recognize what artifact to pick up?
+      `gcr.io/eminent-nation-87317/harbor-builds/harbor-offline-installer-latest.tgz` for development builds
```
andrewtchin: I think we shouldn't use this because we won't be able to tell what version is included in dev builds unless there is some way other than the file name we can use to see the Harbor version

installer/docs/integration/harbor.md
wy65701436: As i know, the current solution of vic is to call it each boot. It's ok to run it every time.

installer/docs/integration/harbor.md
wy65701436: thanks, will add them.

installer/docs/integration/harbor.md
wy65701436: they are pointing to the same build. I do not want to maintain two of them, so let's just add the tgz here.

installer/docs/integration/harbor.md
wy65701436: sure

installer/docs/integration/harbor.md
wy65701436: The start part doesn't depends on admiral.

installer/docs/integration/harbor.md
wy65701436: ok, then let's use the 'latest.build', it contains the build version.

andrewtchin: What is in the file `latest.build` - is that the offline installer or the filename of the latest offline installer tar.gz? That should be documented in here - if it contains a filename, that would be good, but would also require a change to the logic that selects the Harbor artifact to pull in.

wy65701436: The content is the file name, I’m agree that we should define a logic to pick up the latest build by matching installer file instead using the latest.build as a workaround. As the installer file follows the naming convention, the drone job is a sequential number, VIC should use this.

morris-jason: @wy65701436 Last week I update our component inclusion code so that is pulls in the most recent build in `gcr.io/eminent-nation-87317/harbor-builds` based on timestamp, excluding harbor-offline-installer-latest.tgz.

In practice, this seems to work, as the lasest master build picked up `harbor-offline-installer-v1.5.0-build.4252.tgz.` Does this approach work with you? If so, please update the document here.

wy65701436: @morris-jason timestamp is generally fine but you should narrow down files to harbor-offline-installer-${tag}-build.${drone_id}.tgz.

For me, it's better to use the regular expression to match the build number, then to get the highest  number. 

But anyway, we should remove the dependency on either 'latest.build' or 'harbor-offline-installer-latest.tgz'

installer/docs/integration/harbor.md
```diff
@@ -0,0 +1,132 @@
...
+
+- Where will the VIC appliance build obtain component artifacts?
+  Load from harbor offline installer
+  
+    - Link to bucket that will contain development builds: `gcr.io/eminent-nation-87317/harbor-builds`
+    - Link to bucket that will contain tagged builds for releases (including release candidates):
+      `gcr.io/eminent-nation-87317/harbor-releases/${branch_name}`
+    - How will the VIC appliance build recognize what artifact to pick up?
+      `gcr.io/eminent-nation-87317/harbor-builds/master.stable` for development builds, this file holds the URL of stable build.
+      `gcr.io/eminent-nation-87317/harbor-releases/${release_branch}/harbor-offline-installer-${tag}.tgz` for release builds
```
morris-jason: Please place **tagged** release builds in the `gcr.io/eminent-nation-87317/harbor-releases/` bucket directly, not a sub folder.
This means we expect to see:
```
gcr.io/eminent-nation-87317/harbor-releases/harbor-offline-installer-v1.5.0-rc1.tar.gz
gcr.io/eminent-nation-87317/harbor-releases/harbor-offline-installer-v1.5.0.tar.gz
```
but not
```
gcr.io/eminent-nation-87317/harbor-releases/harbor-offline-installer-v1.5.0.build1234.tar.gz
gcr.io/eminent-nation-87317/harbor-releases/harbor-offline-installer-v1.5.0.build.1234.tar.gz
```

installer/docs/integration/harbor.md
```diff
@@ -0,0 +1,132 @@
...
+  Built by `vmware/harbor` https://github.com/vmware/harbor/blob/master/.drone.yml
+
+- Where will the VIC appliance build obtain component artifacts?
+  Load from harbor offline installer
+  
+    - Link to bucket that will contain development builds: `gcr.io/eminent-nation-87317/harbor-builds`
+    - Link to bucket that will contain tagged builds for releases (including release candidates):
+      `gcr.io/eminent-nation-87317/harbor-releases/${branch_name}`
+    - How will the VIC appliance build recognize what artifact to pick up?
+      `gcr.io/eminent-nation-87317/harbor-builds/master.stable` for development builds, this file holds the URL of stable build.
```
morris-jason: Thanks, this correctly addressed https://github.com/vmware/vic-product/issues/1709.

installer/docs/integration/harbor.md
wy65701436: @morris-jason  what's the purpose to do this? as harbor doesn't want to put all the release builds into one bucket(folder), it's hard for us to manage them. This structure was defined in the last release, and already pushed to the community users. 

What's the difficult here? The script should guess the ${release-branch}?

@reasonerjt  what's your thought?

morris-jason: Our release process pulls the latest release candidate build from `gcr.io/eminent-nation-87317/harbor-releases/`, so if you're putting it in ${release-branch}, the script isn't going to find it.

We can change the script, but there are concerns about which harbor to include:
- How do we know which branch to use? Harbor versions are different from ours.
- Should we use the most recent minor version? We didn't do that in vic 1.3.1 when harbor 1.4.0 was released.

>This structure was defined in the last release.

If you look in the bucket, the release candidate builds are in the root of a bucket, not a subfolder. I think Andrew copied them over manually to preserve our original process.

#### Another issue:
We don't want the latest build from `gcr.io/eminent-nation-87317/harbor-releases/${release-branch}` We want the latest *release candidate build*. Right now, if I updated the script to get the latest build from `${release-branch}`, that would be `harbor-offline-installer-v1.5.0-build.3844.tgz.`


I suggest we stick with the current inclusion model for the ova.
Can you update harbor CI to push **only** tagged release versions to the root of the bucket?

morris-jason: >Can you update harbor CI to push only tagged release versions to the root of the bucket?

@reasonerjt @wy65701436 any updates?

morris-jason:
@wy65701436 @reasonerjt closing and merging this, as the harbor v1.5.1 build was uploaded to the proper bucket.

PR #1743, opened 2018-05-14
Nightly ova download on every try

mdharamadas1:
VIC Appliance Checklist:
- [ ] Added tests
- [ ] Tests passing
- [ ] Updated documentation

Fixes #

### DIFF
--- a/tests/vic-product-nightly.sh
+++ b/tests/vic-product-nightly.sh
@@ -52,7 +52,7 @@ echo "Downloading VIC Product OVA build $input..."
 n=0
 until [[ $n -ge 5 ]]
 do
-    wget -P vic-product https://storage.googleapis.com/$GCS_BUCKET/$input && break;
+    wget -o vic-product/$input https://storage.googleapis.com/$GCS_BUCKET/$input && break;
     n=$(($n+1));
     sleep 10;
 done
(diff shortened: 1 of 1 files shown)


tests/vic-product-nightly.sh
```diff
@@ -52,7 +52,7 @@ echo "Downloading VIC Product OVA build $input..."
 n=0
 until [[ $n -ge 5 ]]
 do
-    wget -nc -P vic-product https://storage.googleapis.com/$GCS_BUCKET/$input && break;
+    wget -P vic-product https://storage.googleapis.com/$GCS_BUCKET/$input && break;
```
morris-jason: This will cause downloads of the same file to be downloaded with a numeric suffix if they already exist.
e.g. `ova ova.1 ova.2 ova.3`

Is that what we want here? If we want to download the ova and overwrite it if it exists, we should probably specify `-r`

mdharamadas1: No, but I think `-r` doesn't work either because it does create dir recursively like `storage.googleapis.com/vic-product-ova-releases/vic-v1.2.1-4104e5f9.ova`. So instead I changed it to use `-o` option which will overwrite the file with the same name.

PR #1775, opened 2018-05-22
Create add_custom_ca.md

farflungfish:
Adds a simple procedure to add your custom CA to photon for people who use custom CAs to sign VIC

Documentation update for a solution to a problem I ran into (and spent hours fixing)

### DIFF
--- a/docs/user_doc/vic_app_dev/add_custom_ca.md
+++ b/docs/user_doc/vic_app_dev/add_custom_ca.md
@@ -41,6 +41,7 @@ e.g. docker-entrypoint.sh
 
 ```sh
 #!/bin/sh
+c_rehash
 echo "Injecting CA"
 openssl x509 -in /etc/ssl/certs/root.pem -text >> /etc/pki/tls/certs/ca-bundle.crt
 openssl x509 -in /etc/ssl/certs/root-secondary.pem -text >> /etc/pki/tls/certs/ca-bundle.crt
--- a/docs/user_doc/vic_app_dev/add_custom_ca.md
+++ b/docs/user_doc/vic_app_dev/add_custom_ca.md
@@ -1,0 +1,53 @@
+# Adding a custom Certificate Authority to DCH Photon
+
+If you have deployed the VIC registry with a custom Certificate Authority, this procedure will provide the CA root and other certificates to Photon's trusted root.
+
+You may have seen errors such as the following when attempting to login to your custom registry:
+
+`Error response from daemon: Get https://exampleregistry:443/v2/: x509: certificate signed by unknown authority`
+
+
+**Prerequisites**
+
+- You are using DCH Photon as a Virtual Container Host in a CI or Build / Push setup
+- A custom Certificate Authority was used to generate the VIC SSL certificates
+
+**Procedure**
+
+1. Obtain your Root and/or secondary certificate files
+
+2. Create a Dockerfile which starts from vmware/dch-photon, add commands to:
+    * Copy your root and/or secondary certificates into /etc/ssl/certs
+    * Install openssl-c_rehash
+    * Add in docker-entrypoint.sh (optional if you want to do this on execution of the container)
+
+
+e.g. Dockerfile
+
+```dockerfile
+FROM vmware/dch-photon
+
+# Copy your root and/or secondary certificates into /etc/ssl/certs
+COPY certs/*.crt /etc/ssl/certs/
+
+# Ensure you have openssl-c_rehash installed via tdnf (package manager for photon)
+RUN tdnf install -y openssl-c_rehash
+
+# Create an entrypoint file which hashes the certificates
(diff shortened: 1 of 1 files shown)


andrewtchin:
I think this PR should probably be targeted to `master` branch
What do you think @stuclem?

docs/user_doc/vic_app_dev/add_custom_ca.md
```diff
@@ -0,0 +1,54 @@
...
+e.g. Dockerfile
+
+```dockerfile
+FROM vmware/dch-photon
+
+# Copy your root and/or secondary certificates into /etc/ssl/certs
+COPY certs/*.crt /etc/ssl/certs/
+
+# Ensure you have openssl-c_rehash installed via tdnf (package manager for photon)
+RUN tdnf install -y openssl-c_rehash
```
andrewtchin: What is this used for?

docs/user_doc/vic_app_dev/add_custom_ca.md
```diff
@@ -0,0 +1,54 @@
...
+ADD docker-entrypoint.sh /docker-entrypoint.sh
+```
+
+e.g. docker-entrypoint.sh
+
+```sh
+#!/bin/sh
+c_rehash
+echo "Injecting CA"
+openssl x509 -in /etc/ssl/certs/root.pem -text >> /etc/pki/tls/certs/ca-bundle.crt
```
andrewtchin: Couldn't this be done only once and then these files copied into the container (in the Dockerfile)?
I also think the required filenames and formats and a description of their contents should be included in the `Procedure` section so that the process is clear

stuclem:
First of all, thank you @farflungfish for contributing this! Secondly, I have opened https://github.com/vmware/vic-product/issues/1791 to integrate this new topic into the doc structure. Once approved, we can merge this into the `1.3-docs` branch so that it goes into the 1.3 docs, and then copy it over to master for inclusion in 1.4. I can take care of all of that, once we approve and merge the topic.

stuclem:
@farflungfish, where do you think that this information should reside? My opinion is, rather than keeping this info in a separate topic, it should be integrated into the two existing topics on adding the Registry CA to DCH Photon:

-  [Add the Registry Certificate to a Custom Image](https://vmware.github.io/vic-product/assets/files/html/1.4/vic_app_dev/photon_cert_custom.html)
- [Manually Add the Registry Certificate to a dch-photon Container VM](https://vmware.github.io/vic-product/assets/files/html/1.4/vic_app_dev/photon_cert_manual.html)

Your info can be added as optional steps, along the lines of "If you used a custom CA when you deployed the VCH, then do this...."

What do you think? Thank you again for this contribution, which is the first community contribution to the VIC docs other than fixing the occasional typo!

docs/user_doc/vic_app_dev/add_custom_ca.md
farflungfish: Hi Andrew, unless you "rehash" the CAs, programs such as OpenSSL cannot find any newly added ones. Think of it like a CA index which programs can use. Though oddly I thought ca-bundle.crt within /etc/ssl/tls/ (i cannot recall the exact location at the moment) did the same thing, but apparently not...

docs/user_doc/vic_app_dev/add_custom_ca.md
farflungfish: I believe you need to rehash after the certs are added. I will move this in another commit

docs/user_doc/vic_app_dev/add_custom_ca.md
```diff
@@ -0,0 +1,55 @@
...
+
+e.g. docker-entrypoint.sh
+
+```sh
+#!/bin/sh
+echo "Injecting CA certs"
+openssl x509 -in /etc/ssl/certs/root.pem -text >> /etc/pki/tls/certs/ca-bundle.crt
+openssl x509 -in /etc/ssl/certs/root-secondary.pem -text >> /etc/pki/tls/certs/ca-bundle.crt
+echo "Rehashing new certificates"
+c_rehash
```
farflungfish: With regards to doing this only once and say in the dockerfile at build time, I have found in the past due to AUFS and the layering of docker that files may not change because each layer is read-only and gets a read-write layer added on top - though I have also seen the opposite of this. 

If anything, I prefer to make file changes upon execution as it is more certain and it leaves the container more dynamic in usage.

farflungfish: @stuclem [Add the Registry Certificate to a Custom Image](https://vmware.github.io/vic-product/assets/files/html/1.4/vic_app_dev/photon_cert_manual.html) would be my suggestion. More than welcome, I am flattered, hopefully this saves someone some time down the line!

stuclem: @farflungfish OK, let's put it there. Thanks!

Issue #1612, opened 2018-04-18
Improve basic VCH deployment example


stuclem:
Per chat with @dbarkelew, the basic deployment examples are too basic, given that they lack volumes and registry config. In fact, the dch--photon example is a better minimal example. We should promote the dch-photon example and remove the current basic ones, as they are too basic to be of any use. 

If we can back-fill the change into 1.3 too, that would be ideal.

stuclem:
Here are @dbarkelew 's comments:

------

https://vmware.github.io/vic-product/assets/files/html/1.4/vic_vsphere_admin/deploy_vch_vcenter.html 
If I may, my recommendation is to add “--volume-store datastore:default”. Technically not a minimum install need, but for user experience it would be better if things like “docker run -v myvol:/mnt/data imagename” just work with the default example. I get regular complaints about this cause users don’t know better. Normally working another case. The fix is so fast I don’t always clone a case for these. 

Other suggestion I like to do is “--insecure-registry VIC-appliance-fqdn” and “--insecure-registry VIC-appliance-ip” so users and pull the DCH  image and play with that to push images and things. 

The “--public-network DVSportgroup” is not defined. How would the networking for these examples work out? Maybe I am missing a step. 

One person I worked with liked to look at the vCenter client after every step to see the results of what the configurations were doing to the VM, host and network configurations. I am not sure how to do this to flow well in a document at the moment. Is something to think about to make things more user friendly from a vSphere admin background if a mental connection can be made back to what they already know. Maybe have an example for a VCH or container action, then a screen shot of the vCenter UI circling in red the parts that are a direct result of the configuration in the example. 


------

stuclem:
I have removed the deployment to a cluster example in https://github.com/vmware/vic-product/pull/1722. I retitled the topic on deploying for use with `dch-photon` to **Deploy a Virtual Container Host with vSphere Integrated Containers Registry Access and a Volume Store**. 

To do: update **Verify the Deployment of a VCH** to include pulling `dch-photon` from registry and deploying it.

stuclem:
Fixed in https://github.com/vmware/vic-product/pull/1722. Deserves another read-through before closing.

stuclem:
Updated example is in https://github.com/vmware/vic-product/blob/master/docs/user_doc/vic_vsphere_admin/deploy_vch_dchphoton.md

Further comments from @dbarkelew:

Dani Barkelew [8:49 PM]
[...]

PR #1792, opened 2018-05-25
Further updates to the VCH deployment example

stuclem:
Fixes https://github.com/vmware/vic-product/issues/1612.

@dbarkelew does this better reflect the customer experience? @pdaigle can you please take a look too?
To see the topic in its entirety, rather than the diff, go [here](https://github.com/stuclem/vic-product/blob/basicvch/docs/user_doc/vic_vsphere_admin/deploy_vch_dchphoton.md). Thanks!

### DIFF
--- a/docs/user_doc/vic_app_dev/putting_apps_into_production.md
+++ b/docs/user_doc/vic_app_dev/putting_apps_into_production.md
@@ -2,23 +2,23 @@
 
 vSphere Integrated Containers engine is designed to be a docker API compatible production endpoint for containerized workloads. As such, the design focus is on provisioning containerized applications with optimal isolation, security, data persistence, throughput performance and to take advantage of vSphere capabilities.
 
-vSphere Integrated Containers engine is designed to make existing features of vSphere easy to consume and exploit by providing compatibilty with the Docker image format and Docker client. Inevitably that means that there are some differences between a regular Docker host and a Virtual Container Host (VCH), and between a Linux container and a container VM. Some of those differences are intentional design constraints, such as there being no such thing as a "privileged" container in VIC. Some are because of a lack of functional completeness, while others are outside of the existing scope of the product, such as native support for `docker build`.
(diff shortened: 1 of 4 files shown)


pdaigle:
The only comment I have is about the following statement:

    "[...] and add at least a `default` volume store."

Adding a `default` volume store enables the use of Anonymous Volumes. While Anonymous Volumes are useful for dev & test, I would not recommend systematically enabling them for production use (at the very least, users should understand the consequences of doing so). A better recommendation would be to use a named VolumeStore (e.g. myvolume) to map a vSphere Datastore for consumption via the Docker API.

stuclem:
@pdaigle so, should we add notes recommending against using anonymous volumes in production in the following places?

- [Specify Volume Datastores > Anonymous Volumes](https://vmware.github.io/vic-product/assets/files/html/1.4/vic_vsphere_admin/volume_stores.html#default)
- [Create a Container with a New Anonymous or Named Volume](https://vmware.github.io/vic-product/assets/files/html/1.4/vic_app_dev/using_volumes_with_vic.html#create_container)

If so, how do we reconcile recommending against using anonymous volumes with the statement "some common container images require the presence of a default volume store in order to run" from the first link? 

Also, my understanding is that `dch-photon` creates anonymous volumes, so requires the `default` volume store. Is this indeed the case? 

@dbarkelew and @matthewavery can you chime in here too, please? Thanks!

pdaigle:
The use of anynymous volume is covered in the following section:

[https://vmware.github.io/vic-product/assets/files/html/1.4/vic_app_dev/putting_apps_into_production.html](https://vmware.github.io/vic-product/assets/files/html/1.4/vic_app_dev/putting_apps_into_production.html)

DCH creates two anonymous volumes by default: one for `/certs` and one for `/var/lib/docker`.

image

in `docker inspect`:

image

We can add a recommendation to explicitly create volumes for these which removes the dependency on anonymous volumes:

image

image

Part of this is documented here. We just need to add the '/certs' section with a note that you can also use it to pass certificates in the DCH:

[https://github.com/vmware/vic-product/tree/master/tutorials/dch](https://github.com/vmware/vic-product/tree/master/tutorials/dch)

stuclem:
Further comment from @pdaigle via Slack:

"I just confirmed that DCH fails if no datastore named (default) (i.e. no anonymous) is present.
It’s an OK behavior, I think. We just need to calrify the docs.
SO it does require anonymous by default. Recommendation is to use named volumes unless customer knows what they are doing (some valid use case would be ephemeral docker build hosts for a CI piepline)."

PR #1793, opened 2018-05-25
Example create-vch.json

farflungfish:
Creation of VCHs should be possible by passing in a JSON file to the CLI as well as the API, much like the following example I created. This would make the CLI far more simple to use as well as all of the benefits of storing such configuration as a textual / JSON object.


zjs:
Good idea! This work is tracked by vmware/vic#4622 and vmware/vic#4623.

PR #1807, opened 2018-05-29
Preform a plugin install during the appliance upgrade

### LINKED ISSUE #1702: Automated UI plugin install
**User Statement:**
As a user I want the UI plugin install to be automated.

**Details:**

**Acceptance Criteria:**
- [x] Automated UI plugin install
- [ ] ~Works with `register` API~ - no longer desired.
- [x] Improved logging in `vic-ui` code especially around certs
- [ ] Documentation update
- [ ] Document manual process
- [x] Consider uninstall
- [x] Consider upgrade
- [x] https://github.com/vmware/vic-ui/issues/452

morris-jason:
Fixes #1702

Final change for 1702 to preform a plugin upgrade during the upgrade script.
Merge after 1779.

### DIFF
--- a/installer/build/scripts/upgrade/upgrade.sh
+++ b/installer/build/scripts/upgrade/upgrade.sh
@@ -86,19 +86,71 @@ function usage {
     "
 }
 
-function callRegisterEndpoint {
+# A plugin upgrade is a forced plugin install
+function callPluginUpgradeEndpoint {
+  local preset=$1
+  local vc='{"target":"'"${VCENTER_TARGET}"'","user":"'"${VCENTER_USERNAME}"'","password":"'"${VCENTER_PASSWORD}"'","thumbprint":"'"${VCENTER_FINGERPRINT}"'"}'
+  local plugin='{"preset":"'"${preset}"'","force":true}'
+  local payload='{"vc":'${vc}',"plugin":'${plugin}'}'
+  echo "register payload - ${payload}" | sed -e 's/'${VCENTER_PASSWORD}'/***/g' >> $upgrade_log_file 2>&1
   /usr/bin/curl \
     -k \
     --write-out '%{http_code}' \
     --header "Content-Type: application/json" \
     -X POST \
-    --data '{"target":"'"${VCENTER_TARGET}"'","user":"'"${VCENTER_USERNAME}"'","password":"'"${VCENTER_PASSWORD}"'","externalpsc":"'"${EXTERNAL_PSC}"'","pscdomain":"'"${PSC_DOMAIN}"'"}' \
+    --data "${payload}" \
+    https://localhost:9443/plugin/upgrade
+}
+
+function upgradeAppliancePlugin {
+  # Upgrade the flex client...
+  tab_retries=0
+  max_tab_retries=30 # 5 minutes
+
+  ret=$(callPluginUpgradeEndpoint FLEX)
+  while [[ "$ret" != *"204"* && "$ret" != *"5"* && ${tab_retries} -lt ${max_tab_retries} ]]; do
+    log "Waiting for upgrade appliance flex plugin..."
+    sleep 10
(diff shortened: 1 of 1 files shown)


installer/build/scripts/upgrade/upgrade.sh
```diff
@@ -86,19 +86,71 @@ function usage {
     "
 }
 
+# A plugin upgrade is a forced plugin install
+function callPluginUpgradeEndpoint {
+  local preset=$1
+  local vc='{"target":"'"${VCENTER_TARGET}"'","user":"'"${VCENTER_USERNAME}"'","password":"'"${VCENTER_PASSWORD}"'","thumbprint":"'"${VCENTER_FINGERPRINT}"'"}'
+  local plugin='{"preset":"'"${preset}"'","force":true}'
+  local payload='{"vc":'${vc}',"plugin":'${plugin}'}'
+  echo 'register payload - '${payload}'' | sed -e 's/'${VCENTER_PASSWORD}'/***/g' >> $upgrade_log_file 2>&1
```
zjs: > `echo 'register payload - '${payload}''`

The way quoting works, I think this is equivalent to `echo "register payload - ${payload}"`, which seems slightly clearer.

installer/build/scripts/upgrade/upgrade.sh
```diff
@@ -86,19 +86,71 @@ function usage {
...
+  /usr/bin/curl \
+    -k \
+    --write-out '%{http_code}' \
+    --header "Content-Type: application/json" \
+    -X POST \
+    --data ''${payload}'' \
```
zjs: Are there advantages to this over `--data "${payload}" \`?

installer/build/scripts/upgrade/upgrade.sh
```diff
@@ -516,6 +567,7 @@ function main {
       echo "TLS connection is not secure, unable to proceed with upgrade. Please contact VMware support. Exiting..."
       exit 1
     fi
+    export VCENTER_FINGERPRINT="$(echo ${fingerprint} | awk '{print $2}')"
```
zjs: Perhaps we should double-quote fingerprint for safety? (It shouldn't contain globbing characters, but who knows.)

installer/build/scripts/upgrade/upgrade.sh
```diff
@@ -86,19 +86,71 @@ function usage {
...
+  ret=$(callPluginUpgradeEndpoint H5)
+  while [[ "$ret" != *"204"* && "$ret" != *"5"* && ${tab_retries} -lt ${max_tab_retries} ]]; do
+    log "Waiting for upgrade appliance h5 plugin..."
+    sleep 10
+    let "tab_retries+=1"
+    ret=$(callPluginUpgradeEndpoint H5)
+  done
+
+  if [[ ${tab_retries} -eq ${max_tab_retries} || "$ret" == *"5"* ]]; then
+    log "Failed to upgrade appliance h5 plugin. Check vCenter target settings, or contact VMware support."
```
zjs: Above, we print "WARNING:". Should we do something similar here?

installer/build/scripts/upgrade/upgrade.sh
morris-jason: Should be fine, done!

installer/build/scripts/upgrade/upgrade.sh
morris-jason: No, just sticking with the convention that was already there. I'll switch it, though, it is clearer.

installer/build/scripts/upgrade/upgrade.sh
morris-jason: Hmmm, no. I think if the H5 client upgrade fails, it should fail the upgrade. If flex fails, it could be okay if the vcenter is 6.7.

PR #1831, opened 2018-06-19
Use correct VMDK sizes when building OVA

### LINKED ISSUE #1166: Deploying the OVA with ovftool results in warnings about file size mismatch
**VIC Product version:**

`vic-v1.3.0-rc2-2628-9e2a95b6.ova`

**Deployment details:**

vSphere Client version 6.5.0.10000

**Steps to reproduce:**

Use `ovftool` to deploy the OVA.

**Actual behavior:**

The process ends with:
```
Warning:
 - Wrong file size specified in OVF descriptor for 'vic-disk1.vmdk' (specified: 3276522496, actual 3359524864).
 - Wrong file size specified in OVF descriptor for 'vic-disk2.vmdk' (specified: 274432, actual 274944).
Completed successfully
```

**Expected behavior:**

The process does not include any error messages.

zjs:
The builder OVF used to construct the OVA has hard-coded disk sizes. Correct these to match the final size of the VMDKs to avoid warnings during the OVA deployment process.

VIC Appliance Checklist:
- [X] Up to date with `master` branch
- [ ] Added tests
- [X] Considered impact to upgrade
- [X] Tests passing
- [ ] Updated documentation

Fixes #1166

### DIFF
--- a/installer/build/bootable/build-disks.sh
+++ b/installer/build/bootable/build-disks.sh
@@ -147,11 +147,18 @@ function convert() {
 }
 
 function usage() {
-  echo "Usage: $0 -p package-location -a [create|export] 1>&2"
+  echo "Usage: $0 -p package-location -a [create|export] -i NAME -s SIZE -r ROOT [-i NAME -s SIZE -r ROOT]..."
+  echo "  -p package-location   the working directory to use"
+  echo "  -a action             the action to perform (create or export)"
+  echo "  -i name               the name of an image"
+  echo "  -s size               the size of an image"
+  echo "  -r root               the mount point for the root of an image, relative to the package-location"
+  echo "Example: $0 -p /tmp -a create -i vic-disk1.vmdk -s 6GiB -r mnt/root -i vic-disk2.vmdk -s 2GiB -r mnt/data"
+  echo "Example: $0 -p /tmp -a create -i vic-disk1.vmdk -i vic-disk2.vmdk -s 6GiB -s 2GiB -r mnt/root -r mnt/data"
   exit 1
 }
 
-while getopts "p:a:" flag
+while getopts "p:a:i:s:r:" flag
 do
     case $flag in
 
@@ -161,8 +168,23 @@ do
             ;;
 
         a)
-            # Optional. Offline cache of yum packages
+            # Required. Action: create or export
             ACTION="$OPTARG"
+            ;;
+
+        i)
+            # Required, multi. Ordered list of image names
+            IMAGES+=("$OPTARG")
+            ;;
+
+        s)
+            # Required, multi. Ordered list of image sizes
+            IMAGESIZES+=("$OPTARG")
+            ;;
+
+        r)
+            # Required, multi. Ordered list of image roots
+            IMAGEROOTS+=("$OPTARG")
             ;;
 
         *)
@@ -171,26 +193,9 @@ do
     esac
 done
 shift $((OPTIND-1))
-if [[ -n "$*" || -z "${PACKAGE}" || -z "${ACTION}" ]]; then
-    usage
-fi
-
-# These sizes are minimal for install, since partitions are resized to full disk space after firstboot.
-IMAGESIZES=(
-  "6GiB"
-  "2GiB"
-)
-IMAGES=(
-  "vic-disk1"
-  "vic-disk2"
-)
-IMAGEROOTS=(
-  "${PACKAGE}/mnt/root"
-  "${PACKAGE}/mnt/data"
-)
 
 # check there were no extra args and the required ones are set
-if [[ -n "$*" || -z "${PACKAGE}" || -z "${ACTION}" ]]; then
+if [[ -n "$*" || -z "${PACKAGE}" || -z "${ACTION}" || ${#IMAGES[@]} -eq 0  || ${#IMAGESIZES[@]} -eq 0  || ${#IMAGEROOTS[@]} -eq 0 ]]; then
     usage
 fi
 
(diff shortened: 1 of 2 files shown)


installer/build/bootable/build-main.sh
```diff
@@ -135,7 +135,10 @@ function main {
     log1 "packaging OVA..."
     cp "${DIR}"/config/builder.ovf "${PACKAGE}/vic-${BUILD_OVA_REVISION}.ovf"
     cd "${PACKAGE}"
-    sed -i -e s~--version--~${BUILD_OVA_REVISION}~ vic-${BUILD_OVA_REVISION}.ovf
+    sed -i -e s~--version--~${BUILD_OVA_REVISION}~ \
+           -e "/<File.*vic-disk1.vmdk.*/s|ovf:size=\"[^\"]*\"|ovf:size=\"$(stat --printf="%s" vic-disk1.vmdk)\"|" \
```
hickeng: Please use the same separator characters as above and a space between the filter and the replacement:
`-e "/<File.*vic-disk1.vmdk.*/   s~ovf:size=\"[^\"]*\"~ovf:size=\"$(stat --printf="%s" vic-disk1.vmdk)\"~"`

Add an array of the expected disks and loop over it. Multiple sed passes are a fine cost to pay for increased clarity.

installer/build/bootable/build-main.sh
```diff
@@ -135,10 +135,17 @@ function main {
     log1 "packaging OVA..."
     cp "${DIR}"/config/builder.ovf "${PACKAGE}/vic-${BUILD_OVA_REVISION}.ovf"
     cd "${PACKAGE}"
-    sed -i -e s~--version--~${BUILD_OVA_REVISION}~ vic-${BUILD_OVA_REVISION}.ovf
+    log2 "updating version number"
+    sed -i -e "s|--version--|${BUILD_OVA_REVISION}|" vic-${BUILD_OVA_REVISION}.ovf
+    log2 "updating disk sizes"
+    DISKS=("vic-disk1.vmdk" "vic-disk2.vmdk")
```
morris-jason: All of this logic could all probably go in https://github.com/zjs/vic-product/blob/22cc2ea882424c0b3cf0b80b132d239e8ca73201/installer/build/bootable/build-disks.sh#L208 where disk names are already listed.

zjs: I don't like listing the disk names in three places (the OVF, `build-main.sh`, and `build-disks.sh`), like my current change does.

However, one thing that I like about the current design is that `build-disks.sh` and `build-main.sh` seem to have fairly separate purposes: one creates the VMDKs, while the other bundles them into an OVA.

I'm not sure how important that distinction is in practice, but I worry that moving this logic into `build-disks.sh` blurs the line too much.

Perhaps `build-main.sh` should update the disk sizes, but not have a hardcoded value for `DISKS`. Instead, it could just base the array on the contents of the directory. This is a little fragile (what happens if there's something unexpected in the directory?), but may be better than explicitly listing the names in multiple places. (On the other hand, explicitly referring to them my name makes it easy for someone to grep for them by name and find this logic.)

morris-jason: Fair concerns about blurring the separation of logic between build-main and build-disks.

My original intent was to have all bull attributes (disk size, names, etc) parsed and read from ether a json file or the ovf template in build-main, and passed to the other builder scripts. 

Something similar could be done here - just pass the newly-defined disk array you’ve added in build-main as an argument in calls to build-disk. That preserves the separation of logic while removing one repeated definition of the disks array.

zjs: I did basically what you describe, except to avoid separating the disk names from the other information about them, I moved all three arrays into build-main. I think this is consistent with the idea of eventually reading the values from somewhere (JSON or OVF).

installer/build/bootable/build-disks.sh
```diff
@@ -161,36 +168,34 @@ do
...
-  "vic-disk2"
-)
-IMAGEROOTS=(
-  "${PACKAGE}/mnt/root"
-  "${PACKAGE}/mnt/data"
-)
 
 # check there were no extra args and the required ones are set
-if [[ -n "$*" || -z "${PACKAGE}" || -z "${ACTION}" ]]; then
+if [[ -n "$*" || -z "${PACKAGE}" || -z "${ACTION}" || ${#IMAGES[@]} -eq 0  || ${#IMAGESIZES[@]} -eq 0  || ${#IMAGEROOTS[@]} -eq 0 ]]; then
```
hickeng: From a purely pedantic perspective, these should be tested for equality with each other as well as !0

installer/build/bootable/build-main.sh
```diff
@@ -129,16 +146,22 @@ function main {
...
+    log2 "updating image sizes"
+    for image in "${IMAGEFILES[@]}"
+    do
+        sed -i -e "/<File.*${image}.*/ s|ovf:size=\"[^\"]*\"|ovf:size=\"$(stat --printf="%s" ${image})\"|" vic-${BUILD_OVA_REVISION}.ovf
+    done
     log2 "rebuilding OVF manifest"
-    sha256sum --tag "vic-${BUILD_OVA_REVISION}.ovf" *.vmdk | sed s/SHA256\ \(/SHA256\(/ > "vic-${BUILD_OVA_REVISION}.mf"
-    tar -cvf "${RESOURCE}/vic-${BUILD_OVA_REVISION}.ova" "vic-${BUILD_OVA_REVISION}.ovf" "vic-${BUILD_OVA_REVISION}.mf" *.vmdk
+    sha256sum --tag "vic-${BUILD_OVA_REVISION}.ovf" "${IMAGEFILES[@]}" | sed s/SHA256\ \(/SHA256\(/ > "vic-${BUILD_OVA_REVISION}.mf"
+    tar -cvf "${RESOURCE}/vic-${BUILD_OVA_REVISION}.ova" "vic-${BUILD_OVA_REVISION}.ovf" "vic-${BUILD_OVA_REVISION}.mf" "${IMAGEFILES[@]}"
```
hickeng: This cannot be compressed?

zjs: Filed #1837 to track.

PR #1873, opened 2018-07-03
Prevent password masking from becoming password

hickeng:
Prevent password masking from becoming password

There is a path, likely retry or recoverying but at this time unconfirmed,
by which the password masking can be set as the password. This expressly
checks for the value having been masked and refuses to change the password
to that value.

This is a very focused change to address specifically the observed
behaviour as the implications of a revert or moving the code to other
units are not understood at this time.
Issue was introduced in 3c5cd51a

### DIFF
--- a/installer/build/scripts/systemd/scripts/vic-appliance-environment.sh
+++ b/installer/build/scripts/systemd/scripts/vic-appliance-environment.sh
@@ -14,6 +14,8 @@
 # limitations under the License.
 set -euf -o pipefail
 
+declare -r mask="*******"
+
 umask 077
 
 ENV_FILE="/etc/vmware/environment"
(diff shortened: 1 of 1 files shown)


installer/build/scripts/systemd/scripts/vic-appliance-environment.sh
```diff
@@ -53,7 +55,12 @@ function detectHostname() {
 
 function firstboot() {
   set +e
-  echo "root:$(ovfenv --key appliance.root_pwd)" | chpasswd
+  local tmp="$(ovfenv --key appliance.root_pwd)"
```
zjs: `shellcheck` would [encourage you](https://github.com/koalaman/shellcheck/wiki/SC2155) to split this into two lines to avoid having the return value of `local` hide the return value of your command:
```
local tmp
tmp="$(ovfenv --key appliance.root_pwd)"
```

It's not strictly necessary in this case, and this file has worse issues, but I figured I'd mention it anyway.

hickeng: This is a fair complaint if it weren't for the surrounding `set +e` which negates the value of it. I'm not 100% clear on _why_ those are there, but they are. I can make guesses.

PR #1874, opened 2018-07-03
Prevent password masking from becoming password

hickeng:
There is a path, likely retry or recoverying but at this time unconfirmed,
by which the password masking can be set as the password. This expressly
checks for the value having been masked and refuses to change the password
to that value.

This is a very focused change to address specifically the observed
behaviour as the implications of a revert or moving the code to other
units are not understood at this time.
Issue was introduced in 3c5cd51

(cherry picked from commit 51844757c1dd73af5c8510d0f3e75839ce25340f)

From PR: #1873

### DIFF
--- a/installer/build/scripts/systemd/scripts/vic-appliance-environment.sh
+++ b/installer/build/scripts/systemd/scripts/vic-appliance-environment.sh
@@ -14,6 +14,8 @@
 # limitations under the License.
 set -euf -o pipefail
 
+declare -r mask="*******"
+
 umask 077
 
 ENV_FILE="/etc/vmware/environment"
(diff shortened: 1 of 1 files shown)


zjs:
> (cherry picked from commit 6f82a5d)

This doesn't appear to be the commit that was pushed to `master`, which may make this commit's message confusing to future readers:

```
commit 51844757c1dd73af5c8510d0f3e75839ce25340f (HEAD -> master, origin/master, origin/HEAD)
Author: George Hicken <hickeng@users.noreply.github.com>
Date:   Tue Jul 3 08:51:01 2018 -0700

    Prevent password masking from becoming password (#1873)
    
    There is a path, likely retry or recoverying but at this time unconfirmed,
    by which the password masking can be set as the password. This expressly
    checks for the value having been masked and refuses to change the password
    to that value.
    
    This is a very focused change to address specifically the observed
    behaviour as the implications of a revert or moving the code to other
    units are not understood at this time.
    Issue was introduced in 3c5cd51a
```

PR #1883, opened 2018-07-18
Update branching, cherry-picking, and tagging docs

zjs:
Improve examples and clarify instructions.

---

VIC Appliance Checklist:
- [ ] Added tests
- [x] Considered impact to upgrade
- [x] Tests passing
- [x] Updated documentation

### DIFF
--- a/installer/docs/RELEASE.md
+++ b/installer/docs/RELEASE.md
@@ -38,6 +38,16 @@ After creating a new branch:
    for the `master` branch.
 
 
+### Upgrade
+
+To ensure that future releases will be able to upgrade from the release that is
+being prepared, it is necessary to update the upgrade scripts to list the new
+release. As this process may change over time, this document will not attempt to
+describe the exact steps necessary. Referring to the most recent change that
+added support for upgrade and searching the code for previous version numbers
+may each help identify some of the necessary steps.
+
+
 ### Cherry picking
 
 Commits that need to be pulled into the release should be cherry picked into the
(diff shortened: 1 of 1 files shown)


installer/docs/RELEASE.md
```diff
@@ -1,100 +1,103 @@
...
-git remote update
-git checkout upstream/master
-git tag -a v1.3.0-dev bbbbbbb
-git pubsh upstream v1.3.0-dev
-```
+### Tagging
 
-After the release candidate has passed QA and the team is ready to release, tag the commit in the
-release branch (`v1.2.0`) and push the tag to Github.
+On the `master` branch, tag the commit for the first release candidate. On the
```
hickeng: I'm not sure this should be on `master`. If it's a `Z` release then we're already on a release branch. If it's a `Y` release then we've been creating a release branch from master (we've not yet done an `X` release). We're tagging master with `vX.Y.0-dev` but don't recall that we're also tagging the branch point.

installer/docs/RELEASE.md
```diff
@@ -1,100 +1,103 @@
...
+git checkout upstream/releases/X.Y.Z
+git tag -a vX.Y.Z-rc1
+git push upstream vX.Y.Z-rc1
 ```
 
-### Point releases
+```
+git remote update
+git checkout upstream/master
+git tag -a vA.B.C-dev
```
hickeng: `A.B.C` is a little confusing - I'd suggest either using `X.Y++.0-dev` (or similar) or using specific numeric examples.

installer/docs/RELEASE.md
```diff
@@ -1,100 +1,103 @@
...
+git pubsh upstream vA.B.C-dev
```
hickeng: `git push`.

installer/docs/RELEASE.md
zjs: Yeah, it looks like this section was not completely updated to match the process from the "Branching" section above.

installer/docs/RELEASE.md
zjs: I opted not to use `X.Y++.0` as I _think_ that's a specific case (albeit the most common one). For example, we might decide to skip to `X++.0.0`.

PR #1923, opened 2018-08-13
Fix the certs is signed by nil when FQDN is specified for vic-ip

### LINKED ISSUE #925: dch-photon Automatically Generated Certificates valid for <nil>
**VIC Product version:**
dch-photon:1.13

**Steps to reproduce:**
Tried to run dch-photon for DID setup:
(approach 1) following steps in 
https://vmware.github.io/vic-product/assets/files/html/1.2/vic_app_dev/dchphoton_options.html 
to run dch-photon with TLS Authentication (Automatically Generated Certificates, using VIC IP address as -vic-ip) and docker cp certificates to external host for testing connection
docker run -p 32376:2376 --name my-photon -v myverts:/certs registry:5000/dch-photon:1.13 -tlsverify -vic-ip VIC-IP  

(approach 2) following steps with self-generated certificates (With Remote Verification)
same as above

**Actual behavior:**
both approaches running "docker -H VIC-HOST:32376 --tlsverify info" return
"error during connect: get https://VIC-HOST:VIC-PORT/v1.30/info: x509: certificate is valid for <nil>, not VIC-HOST"
 
**Expected behavior:**
show docker info

Any ideas howto proceed or whats going wrong?

yuyangbj:
VIC Appliance Checklist:
- [ ] Added tests
- [ ] Tests passing
- [ ] Updated documentation

Fixes #925


dinv/dch-photon-1.13/main.go
```diff
@@ -51,6 +51,30 @@ func init() {
...
+		ip, err := getFirstIP("eth0")
+		if err != nil {
+			return nil, err
+		}
+		ip_addr = ip
+	}
+	return ip_addr, nil
+}
+
 func main() {
```
wjun: This api is used in many places, is there any way to remove duplicated codes?

dinv/dch-photon-17.06/main.go
```diff
@@ -79,7 +102,13 @@ func main() {
...
+					ip, err = check_vic_ip(vicIP)
+					if err != nil {
+						ip, err = getFirstIP("eth0")
+						if err != nil {
+							log.Fatal(err.Error())
+						}
+					}
 				} else {
 					ip, err = getFirstIP("eth0")
 					if err != nil {
```
wjun: lines 106-114 seem covered by check_vic_ip api already.

dinv/dch-photon-17.06/main.go
```diff
@@ -108,7 +137,13 @@ func main() {
...
```
wjun: lines 141-148 seem covered by check_vic_ip api already.

dinv/dch-photon-1.13/main.go
hickeng: We would need to move the common code into somewhere accessible during the `docker build` step for each of the images.
The options are:
1. update `build.sh` to copy the file into the build directory before calling build
2. create a common base image with the common code, always build that image first, and use that image as the build-env in the other docker images.

I don't have a strong opinion either way as both add dependencies outside of the build directory and a necessary pre-build operation. (1) is lighter weight.

yuyangbj: Yes, I can do this! But the final result is the files in dcp-photon -xxx will not be a whole docker file. If we leave the code like this, every folder can use docker build to build an image.

hickeng: @frapposelli Do you have an opinion/common approach to the code de-dup issue here? I confess I've not run into it in this specific form before.

zjs:
Could you provide a more complete title and description for this PR? (Perhaps re-using content from your commit messages?) The current title ("Vic ip") isn't very useful for anyone searching or browsing GitHub and the description includes only the default template text.

> Add more information for Group4-Harbor test case 6a501c9 

It seems like this change is separate from the IP-related changes and should go in a separate PR. Having separate pull requests for distinct work items makes it easier to review and cherry-pick changes.

dinv/dch-photon-1.13/main.go
frapposelli: @hickeng one approach I considered was to create common Go packages and fetch them as a dependency during the `docker build` phase (ref. https://github.com/vmware/vic-product/blob/master/dinv/dch-photon-17.06/Dockerfile#L4).

yuyangbj:
@zjs I separate the PRs with following, I will update this PR.

https://github.com/vmware/vic-product/pull/1968

dinv/dch-photon-1.12/main.go
```diff
@@ -114,13 +134,9 @@ func main() {
 			if err := certsExist(files); err != nil {
```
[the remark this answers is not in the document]
yuyangbj: This is not related to this PR, so I think we need to fix them in another PR.

.drone.yml
```diff
@@ -172,17 +172,17 @@ pipeline:
...
       SHELL: /bin/bash
     secrets:
-      - bridge_network
-      - public_network
-      - test_datastore
-      - test_password
-      - test_resource
-      - test_url
-      - test_username
+    - bridge_network
```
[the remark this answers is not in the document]
yuyangbj: I do not need to change this file, so I changed back.

yuyangbj:
@zjs good catch! I leave the main.go in dch-photo-1.12.

dinv/main.go
```diff
@@ -51,6 +51,30 @@ func init() {
 	flag.StringVar(&vicIP, "vic-ip", "", "Set IP for automatic certificate creation")
 }
 
+// This method will check if the vicIP is qualified FQDN or IP address
+// it will return IP address if FQDN is input
+// return nil if non qualified fqdn or ip address is input
+func check_vic_ip(vicIP string) (net.IP, error) {
```
zjs: Can we put this function in a separate file so that it does not need to be duplicated between the two `main.go` files?

yuyangbj: Sure!

dinv/dch-photon-1.12/Dockerfile
```diff
@@ -1,7 +1,9 @@
 # Build dinv in separate container
 FROM golang:1.8 AS build-env
+# copy the non-version specific files first so version specific can overwrite
 ADD . /go/src/dinv
-RUN cd /go/src/dinv && go get -v ./... && go build -o dinv && strip dinv
+ADD ./dch-photon-1.12/* /go/src/dinv/
+RUN rm /go/src/dinv/dch-photon-1.12/main.go && cd /go/src/dinv && go get -v ./... && go build -o dinv && strip dinv
```
hickeng: This functions but I think it may be fragile if any of the other versions ever add version specific source files. This can be simplified to the following:
```Dockerfile
COPY *.go /go/src/dinv/
ADD ./dch-photon-1.12/* /go/src/dinv/
RUN cd /go/src/dinv && go get -v ./... && go build -o dinv && strip dinv
```
i.e. replace `ADD . /go/src/dinv` with `COPY *.go /go/src/dinv` for the non-version specific files and remove the deletion of `dch-photon-1.12/main.go`.

Issue #1967, opened 2018-08-24
Documented instructions for installing the Client Plug-Ins include incorrect bundle version number for 1.4.1 and 1.4.2


zjs:
The [documented instructions](https://vmware.github.io/vic-product/assets/files/html/1.4/vic_vsphere_admin/plugins_vcsa.html) for installing the Client Plug-Ins on a vCenter Server Appliance explicitly state `export VIC_BUNDLE=vic_v1.4.0.tar.gz`. This isn't accurate for releases after 1.4.0.

For product release 1.4.1 and 1.4.2, the correct value is `export VIC_BUNDLE=vic_v1.4.1.tar.gz`.

bug2185982

zjs:
@renmaosheng pointed out that we do have a paragraph that addresses this:

> If you have installed a different version of the appliance, update 1.4.0 to the appropriate version in the command above. You can see the correct version by going to https://vic_appliance_address:9443/files/ in a browser. If the vSphere Integrated Containers appliance was configured to expose the file server on a different port, replace 9443 with the appropriate port.

Given this, I'm lowering the severity, but leaving this open; I think the documentation is technically correct, but it seems like at least three readers (the customer, @dbarkelew, and me) initially missed the paragraph, so there may be room for improvement.

stuclem:
@zjs and @renmaosheng, we previously included `export VIC_BUNDLE=vic_v1.4.x.tar.gz` in the docs, and instructed people to update `x` to whatever minor version they were installing. But, people were also missing this instruction. The point of setting the variables was to make the commands completely cuttable and pastable. I will update this to include a table that includes both 1.4.0 and 1.4.1. For 1.4.3 this is unnecessary, because of the automatic installer.

PR #1971, opened 2018-08-24
Clarified VIC Engine bundle version for plugins

stuclem:
Fixes https://github.com/vmware/vic-product/issues/1967. Clarifies which bundle to use for plug-in installation, and provides copy/pastable commands.


stuclem:
We do not need to state anything for 1.4.3 because plug-in installation/upgrade is automatic.

renmaosheng:
please do a rebase to make the integration test pass.

stuclem:
Merging following approval to do so by @renmaosheng.

Issue #1952, opened 2018-08-22
Ensure that the VIC UI Plugin works when multiple VIC OVAs are present in a vCenter


zjs:
As described in #740, there are cases where a user may deploy multiple VIC OVAs within a single vCenter.

We should ensure that coupling the lifecycle of the VIC UI Plugin with the lifecycle of the VIC OVA does not prevent us from supporting this use case.

Of specific concern is the logic that the UI Plugin uses to find the VIC OVA; it may not behave as expected in the presence of multiple VIC OVAs of different versions.

While `vic-machine-server` is stateless, and therefore any `vic-machine-server` instance of the appropriate version should work with the UI plugin, we should be mindful of other considerations. For example, knowing which instance is used may be important to allow failures to be debugged, using a consistent instance may be necessary to avoid issues related to acceptance of self-signed certificates, and we should be mindful of other differences between VIC OVAs if several are present (power state, network topology, etc.).

In addition to necessary investigation, we should demonstrate that this works via an automated test.

While some of this work is necessary for #1432, it may be possible to defer specific tasks on a case-by-case basis.

stuclem:
There is some discussion of how the plug-in deals with a multi-VC environment in https://github.com/vmware/vic-ui/issues/365. This relates to the old, script-based plug-in installation, rather than to the auto-installation, but might be relevant.

zjs:
Minor note: It is necessary to support downgrade of the vic-ui plugin (to rollback in the case of a failed OVA upgrade), but we probably want that to be somehow _explicit_ so that initializing an old OVA does not cause the downgrade to happen.

zjs:
After discussion with @hickeng, I think there are several things we can do here:

1. Give the user more choice about whether to install the UI plugins.
   * For 1.4.3, provide a checkbox, which is checked by default, on the OVA initialization page: "Install UI Plugin". If this is not checked, do not install/reinstall the UI plugins.
      * This involves changes to the "Getting Started" UI and API, changes to the initialization code which installs the plugin (to make the process conditional), and additional tests.
   * Eventually, provide some sort of warning if downgrade of the UI plugin will occur.
2. Improve the plugin's behavior when multiple OVAs are detected.
   * For 1.4.3, do not attempt to connect to OVAs which are not powered on. Display a clear error message if OVAs are present, but none are powered on.
      * This involves changes to the plugin's OVA lookup logic, changes to the plugin's error handling, and additional tests.
[...]

Issue #1432, opened 2018-02-15
Automate VIC UI plugin install during appliance initialization


andrewtchin:
**User Statement:**
As a user I want the UI plugin install to be automated.

**Details:**
- This should ideally be done by appliance admin API
- During appliance initialization, run vic-ui-linux equivalent to install the plugin
- Install command requires vSphere admin creds, vCenter target, and vCenter fingerprint
- Alert the user to restart vSphere client services or offer to SSH into vCenter to do it (optional)

https://vmware.invisionapp.com/share/T2D63MPMR

andrewtchin:
Previously detailed in https://github.com/vmware/vic/issues/6890. Copying estimate from there

jooskim:
A sample run of `install.sh` assembles and runs the following command:

```
./vic-ui-darwin install --key com.vmware.vic --target $TARGET --user administrator@vsphere.local --configure-ova --type=VicApplianceVM --version 1.3.1.687 --company VMware --url $PLUGIN_ZIP_URL --force --thumbprint C1:6D:3E:41:A5:89:5E:89:FB:DB:82:2F:32:82:8E:54:53:8B:65:34 --server-thumbprint 83:A9:0E:6A:9D:E7:2A:43:B6:70:98:31:86:4A:7D:BC:0A:BE:8A:92 --name vSphere Integrated Containers-H5Client --summary Plugin for vSphere Integrated Containers-H5Client
```

jooskim:
`uninstall.sh`

```
./vic-ui-darwin remove \
     --key com.vmware.vic \
     --target $TARGET \
     --user administrator@vsphere.local \
     --password $PW \
     --thumbprint C1:6D:3E:41:A5:89:5E:89:FB:DB:82:2F:32:82:8E:54:53:8B:65:34
```

`upgrade.sh`

```
./vic-ui-darwin install \
     --force \
     --key com.vmware.vic \
     --target $TARGET \
     --user administrator@vsphere.local \
     --password $PW \
     --configure-ova \
     --type=VicApplianceVM \
     --version 1.3.1.687 \
     --company VMware \
     --url $PLUGIN_ZIP_URL \
     --thumbprint C1:6D:3E:41:A5:89:5E:89:FB:DB:82:2F:32:82:8E:54:53:8B:65:34 \
     --server-thumbprint 83:A9:0E:6A:9D:E7:2A:43:B6:70:98:31:86:4A:7D:BC:0A:BE:8A:92 \
[...]

PR #1908, opened 2018-08-06
Publish artifacts from release branches separately

DanielXiao:
Because builds from all branches are written to the same directory,
it is difficulty to pick up the right build to test against different
release branches.

Publish artifacts from release branches into sub-directories of a
releases directory, so nightly test of a release branch can select the
last build artifact under the subdirectory for testing.

Tags builds are published to a "tags" directory.

VIC Appliance Checklist:
- [ ] Up to date with `master` branch
- [ ] Added tests
- [ ] Considered impact to upgrade
- [ ] Tests passing
- [ ] Updated documentation
- [ ] Impact assessment checklist

If this is a feature or change to existing functionality, consider areas of impact with the [Impact
Assessment Checklist](https://github.com/vmware/vic-product/blob/master/installer/docs/CHANGE.md)

Fixes #

PR #1790, opened 2018-05-24
Add new features for the ui of the vic plugin auto install functionality

lmalvins:
Fix:https://github.com/vmware/vic-ui/issues/465

This PR includes a new modal for plugin installation which includes an ajax call to the server to get the thumbprint. 

Installing for the first time and success message
first_time_success_message

VC validation failed
thumbprint_fail

Wrong credentials
wrong_credentials

Canceling plugin installation
cancel_install

Succes unit test example (jasmine +  karma runner):
success

Failed unit test example (jasmine +  karma runner):
failed

**User Statement:**

As a user, I want to be able to manage VIC UI plugin auto install.
 
**Details:**

**Acceptance Criteria:**
- [x] Add new modal which shows the fingerprint with a cancel and continue button
- [x] Add an ajax call to get the thumbprint from vcenter
- [x] Add call to the API for plugin install
- [x] UX review
- [x] Display success/error messages from auto install
- [x] Alert to restart services
- [x] Add unit testing framework
- [x] Add unit tests

PR #1979, opened 2018-08-27
Merge branch Feature/auto-vic-ui-plugin

wjun:
We switch ui plugin registration from running manual commands of install.sh or upgrade.sh to automation. A set of rest apis are hosted on the VIC appliance fileserver. During the appliance initialization, index.html page will send a post request to the plugin install api, which will prepare the plugin and register the plugin to vCenter.

A thumbprint verification page is shown to enhance security, e.g. a wrong vCenter or a man-in-the-middle attack.

In case there is an old ui plugin in vCenter already, the new VIC appliance will overwrite the old ui plugin and register a new one.

UI plugin registration is also integrated into upgrade.sh, so the UI plugin can be upgraded as well when running upgrade.sh for VIC appliance upgrade.

With UI plugin auto registration, we do not need the manual ui install/upgrade.sh scripts and vic-ui-* commands anymore.

UI plugin registration is set to be installed by default during initialization and upgrade, but users can also disable the installation by uncheck the installation option.

Towards: [vmware/vic-product#1432](https://github.com/vmware/vic-product/issues/1432) [vmware/vic-product#1952](https://github.com/vmware/vic-product/issues/1952)


zjs:
> Add ui plugin installation as an option  c33d39d 

This should be posted for review separately, as a PR targeting the feature branch.

installer/build/scripts/upgrade/upgrade.sh
```diff
@@ -88,19 +88,71 @@ function usage {
...
     https://localhost:9443/register
 }
 
 # Register appliance for content trust
 function registerAppliance {
-
+  log "Registering the appliance in PSC"
   tab_retries=0
   max_tab_retries=30 # 5 minutes
   while [[ "$(callRegisterEndpoint)" != *"200"* && ${tab_retries} -lt ${max_tab_retries} ]]; do
```
zjs: As above.

installer/lib/login.go
```diff
@@ -16,65 +16,91 @@ package lib
...
 
 	ctx, cancel := context.WithTimeout(context.Background(), loginTimeout)
 	loginResponse := make(chan error, 1)
-	var v *validate.Validator
 	var err error
 
 	go func() {
-		v, err = validate.NewValidator(ctx, input)
-		info.Validator = v
+		if info.Thumbprint == "" {
```
zjs: In what case do we expect this value to be empty? Has this block of code been reviewed for potential security issues?

It appears that we're accepting a certificate's thumbprint on behalf of the user:
https://github.com/vmware/vic-product/blob/b65dc1533db3a273f466d74eba62ec3768434161/installer/lib/login.go#L75-L76

It looks like the call to `cert.FromURL` is expected to either return an `err` or set `cert.Err` if the certificate cannot be checked or is invalid (due to an `x509.UnknownAuthorityError` or `x509.HostnameError`). That sounds reasonable, but it's not clear why we want _this_ code doing the validation instead of using a `session.Config` without specifying a thumbprint (which seems like it might be equivalent?).

This must receive appropriate review, if it hasn't already, before merging the feature branch to `master`.

wjun: @morris-jason I think Zach's comment is valid, in that this api will not pass if an empty thumbprint is received?

installer/fileserver/main.go
```diff
@@ -140,14 +140,14 @@ func parseServerConfig(op trace.Operation, conf *serverConfig) {
 // cspMiddleware sets the Content-Security-Policy header to prevent clickjacking
 // https://www.owasp.org/index.php/Content_Security_Policy_Cheat_Sheet#Preventing_Clickjacking
 func cspMiddleware() func(next http.Handler) http.Handler {
-        header := "Content-Security-Policy"
```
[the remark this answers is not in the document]
wjun: Done

installer/build/scripts/upgrade/upgrade.sh
```diff
@@ -46,6 +46,7 @@ DESTROY_ENABLED=""
 MANUAL_DISK_MOVE=""
 EMBEDDED_PSC=""
 INSECURE_SKIP_VERIFY=""
+UPGRADE_UI_PLUGIN=""
```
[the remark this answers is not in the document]
wjun: Done

installer/fileserver/html/.gitignore
```diff
@@ -0,0 +1,41 @@
+# Created by https://www.gitignore.io/api/node
```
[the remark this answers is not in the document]
wjun: Good catch. Remove those test related files during ova build.

installer/build/scripts/upgrade/upgrade.sh
```diff
@@ -626,8 +630,11 @@ function main {
   ### -------------------- ###
   ###  Component Upgrades  ###
   ### -------------------- ###
-  log "\n-------------------------\nStarting VIC UI Plugin Upgrade ${TIMESTAMP}\n"
-  upgradeAppliancePlugin
+  [ -z "${UPGRADE_UI_PLUGIN}" ] && read -p "Upgrade VIC UI Plugin?: (y/n) " UPGRADE_UI_PLUGIN
```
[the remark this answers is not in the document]
wjun: Good point. Done.

tests/test-cases/Group2-Getting-Started/2-01-Getting-Started.robot
```diff
@@ -24,4 +24,4 @@ Verify Getting Started page
     Log To Console  Initializing the OVA using the getting started ui...
     Navigate To Getting Started Page
     Verify Getting Started Page Title
-    Log In And Complete OVA Installation
+    #Log In And Complete OVA Installation
```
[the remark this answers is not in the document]
wjun: There are some issues related to UI page operations this test addresses. @zhoumeina might know how to fix it in the master branch first.

installer/fileserver/html/js/specs/index-handler-spec.js
```diff
@@ -0,0 +1,126 @@
+
```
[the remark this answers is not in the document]
wjun: Done

installer/fileserver/html/package.json
```diff
@@ -0,0 +1,22 @@
+{
```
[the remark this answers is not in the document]
wjun: Done.

installer/build/build-ova.sh
```diff
@@ -57,6 +57,10 @@ do
       VICENGINE="$2"
       shift 2 # past argument
       ;;
+    --vicui)
+      VICUI="$2"
```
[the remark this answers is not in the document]
wjun: Good catch. Done.

installer/build/scripts/upgrade/upgrade.sh
```diff
@@ -88,19 +88,71 @@ function usage {
     "
 }
 
+# A plugin upgrade is a forced plugin install
+function callPluginUpgradeEndpoint {
+  local preset=$1
+  local vc='{"target":"'"${VCENTER_TARGET}"'","user":"'"${VCENTER_USERNAME}"'","password":"'"${VCENTER_PASSWORD}"'","thumbprint":"'"${VCENTER_FINGERPRINT}"'"}'
+  local plugin='{"preset":"'"${preset}"'","force":true}'
+  local payload='{"vc":'${vc}',"plugin":'${plugin}'}'
```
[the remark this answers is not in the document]
wjun: Done

installer/build/scripts/upgrade/upgrade.sh
```diff
@@ -88,19 +88,71 @@ function usage {
...
+    https://localhost:9443/plugin/upgrade
+}
+
+function upgradeAppliancePlugin {
+  # Upgrade the flex client...
+  tab_retries=0
+  max_tab_retries=30 # 5 minutes
+
+  ret=$(callPluginUpgradeEndpoint FLEX)
+  while [[ "$ret" != *"204"* && "$ret" != *"5"* && ${tab_retries} -lt ${max_tab_retries} ]]; do
```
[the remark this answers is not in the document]
wjun: Done

installer/build/scripts/upgrade/upgrade.sh
```diff
@@ -88,19 +88,71 @@ function usage {
...
+
+  if [[ ${tab_retries} -eq ${max_tab_retries} || "$ret" == *"5"* ]]; then
+    log "WARNING: Plugin upgrade failed for the FLEX client. This is expected on vCenter versions 6.7 or higher."
+    log "WARNING: If you expected this to pass on older versions of vSphere, please check your credentials and try again, or contact VMware Support."
+  fi
+
+  # Upgrade the H5 client...
+  tab_retries=0
+  ret=$(callPluginUpgradeEndpoint H5)
+  while [[ "$ret" != *"204"* && "$ret" != *"5"* && ${tab_retries} -lt ${max_tab_retries} ]]; do
```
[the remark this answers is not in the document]
wjun: Done

installer/build/scripts/upgrade/upgrade.sh
```diff
@@ -88,19 +88,71 @@ function usage {
...
+
+  if [[ ${tab_retries} -eq ${max_tab_retries} || "$ret" == *"5"* ]]; then
+    log "Failed to upgrade appliance h5 plugin. Check vCenter target settings, or contact VMware support."
+    exit 1
+  fi
+}
+
 function callRegisterEndpoint {
+  local payload='{"target":"'"${VCENTER_TARGET}"'","user":"'"${VCENTER_USERNAME}"'","password":"'"${VCENTER_PASSWORD}"'","thumbprint":"'"${VCENTER_FINGERPRINT}"'","externalpsc":"'"${EXTERNAL_PSC}"'","pscdomain":"'"${PSC_DOMAIN}"'"}'
+  echo "register payload - ${payload}" | sed -e 's/'${VCENTER_PASSWORD}'/***/g' >> $upgrade_log_file 2>&1
```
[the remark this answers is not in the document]
wjun: Done

installer/scripts/ci-build.sh
```diff
@@ -32,6 +32,9 @@ fi
 if [ -n "${VICENGINE}" ]; then
   OPTIONS="$OPTIONS --vicengine $VICENGINE"
 fi
+if [ -n "${VICUI}" ]; then
+  OPTIONS="$OPTIONS --vicui $VICENGINE"
```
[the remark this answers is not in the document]
wjun: Done

installer/scripts/ci-build.sh
```diff
@@ -57,6 +60,10 @@ if [[ ( "$DRONE_BUILD_EVENT" == "tag" && "$DRONE_TAG" != *"dev"* ) || "$DRONE_BR
     vicengine_release=$(gsutil ls -l "$bucket" | grep -v TOTAL | grep vic_ | sort -k2 -r | (trap '' PIPE; head -1) | xargs | cut -d ' ' -f 3 | sed 's/gs:\/\//https:\/\/storage.googleapis.com\//')
     OPTIONS="$OPTIONS --vicengine ${vicengine_release:?Unable to find an appropriate VIC Engine build. Is '"'$DRONE_BRANCH'"' a valid vmware/vic branch?}"
   fi
+  if [ -z "${VICUI}" ]; then
+    vicui_release=$(gsutil ls -l "gs://vic-ui-releases" | grep -v TOTAL | grep vic_ | sort -k2 -r | (trap '' PIPE; head -1) | xargs | cut -d " " -f 3 | sed 's/gs:\/\//https:\/\/storage.googleapis.com\//')
```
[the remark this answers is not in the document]
wjun: Done

installer/lib/login.go
morris-jason: @zjs @wjun the thumbprint should *always* be supplied by the calling user - both when that call is to the API or from the Getting Started Page's POST to the /login page. I think this code was here during the plugin development process when the thumbprint input form was not yet finished and we needed a mechanism to complete the plugin install without a thumbprint.
This section (L60 - L77) should be removed in favor of parameter verification before this goroutine is launched.

wjun: @morris-jason Thanks for the information. I just refactored the codes with thumbprint empty check.

installer/fileserver/main.go
zjs: The feedback from https://github.com/vmware/vic-product/pull/1979/commits/52430cddeff60119b4afee9763aa9a8bf22d1b91#r213383274 was not correctly addressed; this change was incorporated into "Implement fileserver thumbprint verification (#1790)" _not_ "Automate VIC UI Plugin Install (#1758 #1766 #1785 #1799)".

This feedback must be addressed before merging the feature branch to `master`.

installer/build/scripts/upgrade/upgrade.sh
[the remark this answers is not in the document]
zjs: Thanks!

installer/build/scripts/upgrade/upgrade.sh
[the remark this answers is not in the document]
zjs: Thanks!

installer/build/scripts/upgrade/upgrade.sh
```diff
@@ -658,7 +665,11 @@ function finish() {
   if [ "$rc" -eq 0 ]; then
     log ""
     log "-------------------------"
-    log "Upgrade completed successfully. Exiting."
+    if [ "$UPGRADE_UI_PLUGIN" == "y" ]; then
+      log "Upgrade completed successfully. Exiting. Please logout and login vCenter Server twice to view the upgraded ui plugin."
```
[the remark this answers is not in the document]
zjs: To clarify this question: why does the user need to log out and log in _twice_?

The [previous instructions](https://github.com/vmware/vic-ui/pull/577/files#diff-ba7553843e2b38ed112b326468c007d7L271) only instructed the user to log out and log once:

> ... log out and log back in again ...

installer/fileserver/routes/index.go
```diff
@@ -60,6 +60,8 @@ func (i *IndexHTMLRenderer) IndexHandler(resp http.ResponseWriter, req *http.Req
 		if err := indexFormHandler(op, req, html); err != nil {
 			op.Errorf("Install failed: %s", err.Error())
 			html.InitErrorFeedback = fmt.Sprintf("Installation failed: %s", err.Error())
+		} else if req.FormValue("needuiplugin") == "true" {
+			html.InitSuccessFeedback = "Installation successful. Refer to the Post-install and Deployment tasks below. Please logout and login vCenter Server twice to view the latest ui plugin."
```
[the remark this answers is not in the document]
zjs: To clarify this question: why does the user need to log out and log in _twice_?

The [previous instructions](https://github.com/vmware/vic-ui/pull/577/files#diff-ba7553843e2b38ed112b326468c007d7L271) only instructed the user to log out and log once:

> ... log out and log back in again ...

tests/test-cases/Group2-Getting-Started/2-01-Getting-Started.robot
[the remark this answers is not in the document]
zjs: I don't think we should merge to `master` without fixing the very minimal automated testing we have.

installer/build/scripts/upgrade/upgrade.sh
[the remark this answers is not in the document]
zjs: I don't think "Make ui plugin installation optional" is the correct commit to fix this in (as it is unrelated to making the installation optional). This should either be fixed in a new commit or corrected in the problematic commit ("Automate VIC UI Plugin Install").

installer/build/scripts/upgrade/upgrade.sh
```diff
@@ -140,10 +143,12 @@ function upgradeAppliancePlugin {
 
 function callRegisterEndpoint {
   local payload='{"target":"'"${VCENTER_TARGET}"'","user":"'"${VCENTER_USERNAME}"'","password":"'"${VCENTER_PASSWORD}"'","thumbprint":"'"${VCENTER_FINGERPRINT}"'","externalpsc":"'"${EXTERNAL_PSC}"'","pscdomain":"'"${PSC_DOMAIN}"'"}'
-  echo "register payload - ${payload}" | sed -e 's/'${VCENTER_PASSWORD}'/***/g' >> $upgrade_log_file 2>&1
+  echo "register payload - ${payload}" | sed -e 's/"${VCENTER_PASSWORD}"/***/g' >> $upgrade_log_file 2>&1
```
zjs: This no longer censors the password at all; it just censors the literal string `${VCENTER_PASSWORD}`.

(Additionally, this is not the right commit to make this change in.)

installer/build/scripts/upgrade/upgrade.sh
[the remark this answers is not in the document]
zjs: I don't think "Make ui plugin installation optional" is the correct commit to fix this in (as it is unrelated to making the installation optional). This should either be fixed in a new commit or corrected in the problematic commit ("Automate VIC UI Plugin Install").

installer/build/scripts/upgrade/upgrade.sh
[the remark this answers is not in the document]
zjs: I don't think "Make ui plugin installation optional" is the correct commit to fix this in (as it is unrelated to making the installation optional). This should either be fixed in a new commit or corrected in the problematic commit ("Automate VIC UI Plugin Install").

installer/scripts/ci-build.sh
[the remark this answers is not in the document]
zjs: I don't think "Make ui plugin installation optional" is the correct commit to fix this in (as it is unrelated to making the installation optional). This should either be fixed in a new commit or corrected in the problematic commit ("Automate VIC UI Plugin Install").

installer/scripts/ci-build.sh
```diff
@@ -61,8 +61,13 @@ if [[ ( "$DRONE_BUILD_EVENT" == "tag" && "$DRONE_TAG" != *"dev"* ) || "$DRONE_BR
     OPTIONS="$OPTIONS --vicengine ${vicengine_release:?Unable to find an appropriate VIC Engine build. Is '"'$DRONE_BRANCH'"' a valid vmware/vic branch?}"
   fi
   if [ -z "${VICUI}" ]; then
-    vicui_release=$(gsutil ls -l "gs://vic-ui-releases" | grep -v TOTAL | grep vic_ | sort -k2 -r | (trap '' PIPE; head -1) | xargs | cut -d " " -f 3 | sed 's/gs:\/\//https:\/\/storage.googleapis.com\//')
-    OPTIONS="$OPTIONS --vicui $vicui_release"
+    if [[ "$DRONE_BUILD_EVENT" != "tag" || "$DRONE_TAG" == *"dev"* ]]; then
+      bucket="gs://vic-ui-builds/$DRONE_BRANCH"
```
zjs: Has this change been tested? Based on code inspection, I think this would break when merged to the release branch; I think you'll need to change the way vmware/vic-ui publishes builds for this to work. (Currently, its release branch builds are not published to a subdirectory.) I'd recommend doing so as a separate PR.

Additionally, I don't think "Make ui plugin installation optional" is the correct commit to fix this in (as it is unrelated to making the installation optional). This should either be fixed in a new commit or corrected in the problematic commit ("Automate VIC UI Plugin Install").

installer/build/build-ova.sh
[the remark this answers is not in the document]
zjs: I don't think "Make ui plugin installation optional" is the correct commit to fix this in (as it is unrelated to making the installation optional). This should either be fixed in a new commit or corrected in the problematic commit ("Automate VIC UI Plugin Install").

installer/fileserver/html/js/specs/index-handler-spec.js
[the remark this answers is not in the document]
zjs: I don't think "Make ui plugin installation optional" is the correct commit to fix this in (as it is unrelated to making the installation optional). This should either be fixed in a new commit or corrected in the problematic commit ("Implement fileserver thumbprint verification").

installer/build/bootable/build-main.sh
```diff
@@ -86,6 +86,11 @@ function build_app {
                 SOURCE=$(jq '.['$LINE_NUM'] | .source' "${MANIFEST}" | tr -d '"')
                 DESTINATION=$(echo "${ROOT}/$(cat "${MANIFEST}" | jq '.['$LINE_NUM'] | .destination')" | tr -d '"' )
                 mkdir -p "$(dirname "$DESTINATION")" && cp -R $SOURCE "$DESTINATION"
+                if [[ "$DESTINATION" == *"fileserver/html"* ]]; then
```
zjs: Currently `build-main.sh` does not know anything about the contents of the disks it builds; all of that information should come from the manifest file.

I think we should fix this by correcting the manifest or moving files to different directories, not adding a hack here to delete specific files.

installer/lib/login.go
zjs: I don't think "Make ui plugin installation optional" is the correct commit to fix this in (as it is unrelated to making the installation optional). This should either be fixed in a new commit or corrected in the problematic commit ("Automate VIC UI Plugin Install").

installer/scripts/ci-build.sh
zjs: If you need help making this change, I can take care of it tomorrow. Alternatively, @DanielXiao may be able to help as it's very similar to #1908, but for the vmware/vic-ui repository.

installer/lib/login.go
wjun: Fixed.

installer/build/scripts/upgrade/upgrade.sh
```diff
@@ -88,19 +88,75 @@ function usage {
     "
 }
 
+# A plugin upgrade is a forced plugin install
+function callPluginUpgradeEndpoint {
+  local preset=$1
+  local vc='{"target":"'"${VCENTER_TARGET}"'","user":"'"${VCENTER_USERNAME}"'","password":"'"${VCENTER_PASSWORD}"'","thumbprint":"'"${VCENTER_FINGERPRINT}"'"}'
+  local plugin='{"preset":"'"${preset}"'","force":true}'
+  local payload='{"vc":"${vc}","plugin":"${plugin}"}'
+  echo "register payload - ${payload}" | sed -e 's/'${VCENTER_PASSWORD}'/***/g' >> $upgrade_log_file 2>&1
```
zjs: It seems like this has the previously-mentioned issue where some choices of password can cause a variety of very strange effects here.

As this is a potential security issue, I think we should address this before merging to `master`.

installer/build/scripts/upgrade/upgrade.sh
```diff
@@ -88,19 +88,75 @@ function usage {
...
+
+  if [[ ${tab_retries} -eq ${max_tab_retries} || "$ret" == *"5"* ]]; then
+    log "Failed to upgrade appliance h5 plugin. Check vCenter target settings, or contact VMware support."
+    exit 1
+  fi
+}
+
 function callRegisterEndpoint {
+  local payload='{"target":"'"${VCENTER_TARGET}"'","user":"'"${VCENTER_USERNAME}"'","password":"'"${VCENTER_PASSWORD}"'","thumbprint":"'"${VCENTER_FINGERPRINT}"'","externalpsc":"'"${EXTERNAL_PSC}"'","pscdomain":"'"${PSC_DOMAIN}"'"}'
+  echo "register payload - ${payload}" | sed -e 's/'${VCENTER_PASSWORD}'/***/g' >> $upgrade_log_file 2>&1
```
zjs: As above.

tests/test-cases/Group2-Getting-Started/2-01-Getting-Started.robot
```diff
@@ -15,11 +15,13 @@
...
 Test Teardown  Close All Browsers
 
 *** Test Cases ***
 Verify Getting Started page
+    Log To Console  Initializing the OVA using the getting started ui...
     Navigate To Getting Started Page
-    Verify Getting Started Page Title
\ No newline at end of file
+    Verify Getting Started Page Title
+    #Log In And Complete OVA Installation
```
zjs: I don't think we should merge to `master` without having some very basic automated integration testing. With this line commented out, it seems like our only testing is via scenario tests.

(Previous discussion: https://github.com/vmware/vic-product/pull/1979#discussion_r213756080)

installer/fileserver/html/js/fixtures/index.html
```diff
@@ -0,0 +1,96 @@
+<div id="main-container">
```
zjs: This duplication should be addressed for 1.4.3, but I think this can happen after we merge to `master`.

(Previous discussion: https://github.com/vmware/vic-product/pull/1979/commits/bbfe684912a9c3a5c7266b8bb79aa8cdb1032ff1#r213637735)

installer/build/bootable/build-main.sh
zjs: I think a better short-term hack would be to update the manifest to explicitly whitelist files/directories under `fileserver/html` that we wish to include, instead of deleting files here.

(Previous discussion: https://github.com/vmware/vic-product/pull/1979/commits/bbfe684912a9c3a5c7266b8bb79aa8cdb1032ff1#r213761707)

installer/build/scripts/provisioners/provision_fileserver.sh
```diff
@@ -19,17 +19,52 @@ FILES_DIR="${DATA_DIR}/files"
...
+cd ${VIC_BIN_ROOT}ui/vsphere-client-serenity/com.vmware.vic.ui-${VIC_UI_VER_STRING}
+sed -i "s/Flex Client Plugin for vSphere Integrated Containers Engine/vSphere Client Plugin for vSphere Integrated Containers Engine \(v${VIC_ENGINE_VER_STRING}\)/" plugin-package.xml 
+zip -9 -r ${VIC_BIN_ROOT}ui/vsphere-client-serenity/com.vmware.vic.ui-${VIC_UI_VER_STRING}.zip ./*
+cd ${TMP_FOLDER}
+
+# update plugin-manifest
+sed -i "s/summary=.*/summary=\"vSphere Client Plugin for vSphere Integrated Containers Engine (v${VIC_ENGINE_VER_STRING})\"/" ${VIC_BIN_ROOT}ui/plugin-manifest 
+
+echo "version from the vic-ui repo is:  ${VIC_UI_VER_STRING}"
+echo "version from vic-appliance is:    ${VIC_ENGINE_VER_STRING}"
```
zjs: > I don't understand exactly what these changes are doing. However, this line doesn't seem to make any sense (the version of the appliance isn't going to be the same as the version of engine).
> 
> Maybe it's just a mistake in the string, but I think these changes warrant additional review and manual testing. (E.g., we should look at what happens when building an OVA using the feature branch components and verify that we get the expected version numbers.)
>
> This does not need to be resolved before merging the feature branch to `master`. However, in that case an issue should be filed to track this work before the merge occurs.

(Previous discussion: https://github.com/vmware/vic-product/pull/1979/commits/5b5bb71d0af40a956f62089f385f233f83e95eea#r213408342)

installer/scripts/ci-build.sh
```diff
@@ -57,6 +60,15 @@ if [[ ( "$DRONE_BUILD_EVENT" == "tag" && "$DRONE_TAG" != *"dev"* ) || "$DRONE_BR
     vicengine_release=$(gsutil ls -l "$bucket" | grep -v TOTAL | grep vic_ | sort -k2 -r | (trap '' PIPE; head -1) | xargs | cut -d ' ' -f 3 | sed 's/gs:\/\//https:\/\/storage.googleapis.com\//')
     OPTIONS="$OPTIONS --vicengine ${vicengine_release:?Unable to find an appropriate VIC Engine build. Is '"'$DRONE_BRANCH'"' a valid vmware/vic branch?}"
   fi
+  if [ -z "${VICUI}" ]; then
+    if [[ "$DRONE_BUILD_EVENT" != "tag" || "$DRONE_TAG" == *"dev"* ]]; then
+      bucket="gs://vic-ui-builds/$DRONE_BRANCH"
```
zjs: > Based on code inspection, I think this would break when merged to the release branch; I think you'll need to change the way vmware/vic-ui publishes builds for this to work. (Currently, its release branch builds are not published to a subdirectory.) It is necessary to do so as a separate PR (as that will need to merge first, then this, followed by the feature branch merges in vic-ui and vic).

> If you need help making this change, I can take care of it tomorrow. Alternatively, @DanielXiao may be able to help as it's very similar to #1908, but for the vmware/vic-ui repository.


(Previous discussion: https://github.com/vmware/vic-product/pull/1979/commits/5b5bb71d0af40a956f62089f385f233f83e95eea#r213759971)

installer/fileserver/routes/index.go
```diff
@@ -60,6 +60,8 @@ func (i *IndexHTMLRenderer) IndexHandler(resp http.ResponseWriter, req *http.Req
 		if err := indexFormHandler(op, req, html); err != nil {
 			op.Errorf("Install failed: %s", err.Error())
 			html.InitErrorFeedback = fmt.Sprintf("Installation failed: %s", err.Error())
+		} else if req.FormValue("needuiplugin") == "true" {
+			html.InitSuccessFeedback = "Installation successful. Refer to the Post-install and Deployment tasks below. All vSphere Client users must log out and log back in again twice to see the vSphere Integrated Containers plug-in."
```
zjs: We should file an issue to investigate further; logging out twice should not be required.

(Previous discussion: https://github.com/vmware/vic-product/pull/1979#discussion_r213386720)

zjs:
Rebasing again on `master` should fix some of the CI failures.

installer/scripts/ci-build.sh
wjun: Thanks Yifeng to address it at https://github.com/vmware/vic-ui/pull/582

installer/build/scripts/upgrade/upgrade.sh
wjun: Done

installer/build/scripts/upgrade/upgrade.sh
wjun: Done

installer/build/scripts/upgrade/upgrade.sh
```diff
@@ -89,19 +89,75 @@ function usage {
     "
 }
 
+# A plugin upgrade is a forced plugin install
+function callPluginUpgradeEndpoint {
+  local preset=$1
+  local vc='{"target":"'"${VCENTER_TARGET}"'","user":"'"${VCENTER_USERNAME}"'","password":"'"${VCENTER_PASSWORD}"'","thumbprint":"'"${VCENTER_FINGERPRINT}"'"}'
+  local plugin='{"preset":"'"${preset}"'","force":true}'
+  local payload='{"vc":"${vc}","plugin":"${plugin}"}'
+  echo "register payload - ${payload}" | sed -e 's/'"${VCENTER_PASSWORD}"'/***/g' >> $upgrade_log_file 2>&1
```
zjs: > `sed -e 's/'"${VCENTER_PASSWORD}"'/***/g'`

This is better, but I still think there are some issues with it. As a silly example, a password of `register.*` will cause the whole line to be replace with `***`.

One solution might be to use something like [`sed -e 's/[]\/$*.^[]/\\&/g'`](https://stackoverflow.com/a/2705678) to escape any special characters in `${VCENTER_PASSWORD}` before passing it to `sed`.

This would give us something like:

> `sed -e 's/"password":"'"$(echo "${VCENTER_PASSWORD}" | sed -e 's/[]\/$*.^[]/\\&/g')"'/***/g'`

Another solution might be to use a different command to perform the substitution; `sed` may be overkill for this.

A third solution would be to refactor the code to avoid the need for the replacement entirely; we put the password into the string right before working hard to remove it.

wjun: According to vCenter password policy, register.* is an invalid password.

wjun: Removed sed or any replace approach, and use string without password variable instead for log only now.

installer/build/bootable/build-main.sh
wjun: There are quite a few files and directories under html, compared with the ones we want to delete. The whitelist will always ask UI developers to maintain the changes to whitelist which is a burden to them, while deletion with * will not introduce any extra work.

wjun: Discussion is tracked below with short term solution to deleting these directories during build.

installer/fileserver/html/js/fixtures/index.html
wjun: #1996  has been filed to track, but I would suggest to review and prioritize to see if this can be addressed in release 1.4.3.

installer/build/scripts/provisioners/provision_fileserver.sh
wjun: Good catch. This should be vic-engine. Fixed.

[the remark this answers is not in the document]
wjun: It should be vic-engine. Fixed.

installer/fileserver/html/js/fixtures/index.html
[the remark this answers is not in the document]
wjun: #1996  has been filed to track, but I would suggest to review and prioritize to see if this can be addressed in release 1.4.3.

tests/test-cases/Group2-Getting-Started/2-01-Getting-Started.robot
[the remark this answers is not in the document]
wjun: We have fixed up ui plugin automation through api tests. This test clicks VIC appliance initialization UI pages and a few pages and is out of sync with current web page css contents. We will address this test case tracked by https://github.com/vmware/vic-product/issues/1997

PR #1985, opened 2018-08-29
Fix scenario test failures

DanielXiao:
Fix a couple of scenario test failures.


zjs:

Issue #1933, opened 2018-08-15
CVE-2018-3620 - L1 terminal fault vulnerability


hickeng:
@hickeng commented on [Wed Aug 15 2018](https://github.com/vmware/vic/issues/8211)

Update PhotonOS kernel packages to address 
https://www.vmware.com/security/advisories/VMSA-2018-0021.htm

CVE-2018-3620

Ongoing industry work to address specter/meltdown

Kernel version is expected to be `4.4.148-1` but should be confirmed before closing.

renmaosheng:
from https://bugzilla.eng.vmware.com/show_bug.cgi?id=2186410#c10, the 4.4.152-1.ph1 is ready, thanks.

stuclem:
Moved up to Resolved issues in https://github.com/vmware/vic-product/releases/tag/v1.4.3

PR #1992, opened 2018-08-30
[WIP]Build OS packages from local yum repo

DanielXiao:
To keep consistent with vic buld and control installed
package version, use local yum repository instead of
the remote one.

The local yum repo is updated with below image and
includes CVE-2018-3620 fix with linux 4.4.152-1.ph1.

gcr.io/eminent-nation-87317/local-repo:1.7

VIC Appliance Checklist:
- [ ] Added tests
- [ ] Tests passing
- [ ] Updated documentation

Related: #1933, vmware/vic#8211

### DIFF
--- a/installer/build/bootable/build-base.sh
+++ b/installer/build/bootable/build-base.sh
@@ -35,13 +35,11 @@ function set_base() {
   mkdir -p "${rt}/etc/yum.repos.d/"
   rm /etc/yum.repos.d/{photon,photon-updates}.repo
   cp "${DIR}"/repo/*-remote.repo /etc/yum.repos.d/
-  # TODO: Use local yum repo in CI
-  # if [[ $DRONE_BUILD_NUMBER && $DRONE_BUILD_NUMBER > 0 ]]; then
-  #   mkdir -p /etc/yum.repos.d.old/
-  #   mv /etc/yum.repos.d/* /etc/yum.repos.d.old/
-  #   cp repo/*-local.repo /etc/yum.repos.d/
-  # fi
   cp -a /etc/yum.repos.d/ "${rt}/etc/"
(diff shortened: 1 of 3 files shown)


installer/build/bootable/build-base.sh
```diff
@@ -35,13 +35,11 @@ function set_base() {
   mkdir -p "${rt}/etc/yum.repos.d/"
   rm /etc/yum.repos.d/{photon,photon-updates}.repo
   cp "${DIR}"/repo/*-remote.repo /etc/yum.repos.d/
-  # TODO: Use local yum repo in CI
-  # if [[ $DRONE_BUILD_NUMBER && $DRONE_BUILD_NUMBER > 0 ]]; then
-  #   mkdir -p /etc/yum.repos.d.old/
-  #   mv /etc/yum.repos.d/* /etc/yum.repos.d.old/
```
zjs: I think the old configuration was being saved in the commented-out code because we need to restore it at the end to ensure that yum is configured to use an accessible repo when the OVA is deployed. There's probably other ways to accomplish this, but the goal is to use the local cache when _building_ the OVA, but public repos when running `tdnf` from a deployed OVA (as a customer would).

DanielXiao: I didn't change any thing in ${rt}/etc/yum.repos.d/, repo file should be the same as before. I just use local cache during installing packages. A customer deployed OVA still use *-remote.repo.

DanielXiao: Below command copies all repo files to target VIC vApp root disk, I only modifies repo files of the build machine OS to use local repo. The build machine is deleted afterward so we don't need to restore it.

cp -a /etc/yum.repos.d/ "${rt}/etc/"

DanielXiao:
This is already addressed.

PR #2017, opened 2018-09-04
Quote potentially empty param to nimbus keyword

hickeng:
If omitting the ESX and vCenter version parameters to the Nimbus Testbed
Deploy keyword we get arguments to the underlying script being position
shifted and causing confusing output/failures.
This quotes the provided strings to ensure that even if empty we keep
the argument positions when calling the underlying scripts.

Towards #1957


hickeng:
I'm sure the other repos suffer similar issues, but just fixing it where it's been observed with the expectation that we move the bulk of this infrastructure keyword logic to vic-tools for maintenance in a single location.

zjs:

hickeng:
@zjs shouldn't be needed - if the job is invoked with the variables set then this has no functional effect. Added this because tracking down what was going on was confusing and I prefer good/obvious error messages. For 1.4.3 it's likely just noise so long as the jenkins jobs are not modified to remove the version parameters

PR #2023, opened 2018-09-06
Correct vic_*.tar.gz file lookup

wjun:
Since we have both vic_*.tar.gz and vic_ui_*.tar.gz under files
directory, we need to exclude vic_ui_*.tar.gz


renmaosheng:
please add more content in the commit message to describe the issue we want to fix. thanks..

installer/fileserver/main.go
```diff
@@ -129,7 +129,7 @@ func parseServerConfig(op trace.Operation, conf *serverConfig) {
 
 	// get the fileserver vic tar location
 	filepath.Walk("/opt/vmware/fileserver/files/", func(path string, f os.FileInfo, err error) error {
-		if strings.HasSuffix(path, ".tar.gz") {
+		if strings.HasSuffix(path, ".tar.gz") && !strings.Contains(path, "vic_ui") {
```
DanielXiao: Maybe check file name not start with "vic_ui" is more accurate.

PR #2024, opened 2018-09-06
Change vic ui version string to remove prefix

wjun:
Some versions include '-' which will fail


zjs:

PR #2026, opened 2018-09-06
refactor test folder

zhoumeina:
refine vic-product install files code structure. Extract test folder to avoid javascript errors.


zjs:

wjun:
@zjs Since there is no functional change for this PR, I would not suggest to cherry pick into 1.4.3 branch at the RC stage.

PR #2038, opened 2018-09-07
Adding the source files for the networking and roles graphics

stuclem:
Keeping an archive of the graphic source files here, because the VIC doesn't fit in with the art department's usual archiving system.


stuclem:
Thanks @vidya-v !

PR #2040, opened 2018-09-10
Increase Nimbus testbed resources

DanielXiao:
Increase CPU and Memory for Nimbus testbed to get better
infrastructure performance and reduce intermittent errors.

VIC Appliance Checklist:
- [ ] Added tests
- [ ] Tests passing
- [ ] Updated documentation

Fixes #


zjs:
Why is the lease duration being extended?

zjs:
> Why is the lease duration being extended?

The upgrade tests sometimes take longer than 6 hours to run.

tests/resources/OVA-Util.robot
```diff
@@ -302,7 +302,7 @@ Setup Simple VC And Test Environment
     [Timeout]    110 minutes
     Run Keyword And Ignore Error  Nimbus Cleanup  ${list}  ${false}
     # setup nimbus testbed
-    ${esx1}  ${esx2}  ${vc}  ${esx1-ip}  ${esx2-ip}  ${vc-ip}=  Create a Simple VC Cluster  ha-datacenter  cls  2
+    ${esx1}  ${esx2}  ${vc}  ${esx1-ip}  ${esx2-ip}  ${vc-ip}=  Create a Simple VC Cluster  ha-datacenter  cls  1
```
renmaosheng: it seems we can't just reduce the esx number, some function requires esx2 as parameter.

DanielXiao: Done

renmaosheng:
while we are debugging the ova upgrade tests, we found the environment has two esx hosts, but we only using datastore1, so that means two OVAs are deployed to single ESX host, it causes the ssh to ova extremely slow to perform the upgrade, adjusting the resource is a must. after the fix, we can see that all upgrade tests passed: https://vic-jenkins.eng.vmware.com/view/Scenario/job/vic-product-release-scenario-test/38/ .
From long term perspective, we should set up a shared storage Nimbus testbed to allow OVA deployed to different ESX hosts.

zjs:
What is the status of this PR?

PR #2041, opened 2018-09-10
Fix Group9-VIC-UI test failure

wjun:
Comment out flex plugin uninstall case because flex installation
on vCenter 6.7 is not supported currently.


zjs:
I think we should make this logic conditional instead of commenting it out. It seems like this change will cause us to stop uninstalling on all versions.

Because this is a scenario test, it will normally run against a clean Nimbus environment each time. However, this test would now leave things in a different state if run against a developer's own environment (e.g., for debugging a failure).

wjun:
@zjs Since we may not support flex plugin in the future with a great possibility, I will not install flex in the setup stage. Actually, currently, it will fail on vCenter 6.5/6.7, which nightly is running on.

Issue #2050, opened 2018-09-11
Upgrade from VIC 1.3.1 to 1.4.3 failed because h5 and flex plugin cannot be upgraded


jitinkumar2018:
Bug Report Basic Information
REQUIRED:

vCenter Server version: 6.5 U2
Embedded or external PSC: No
Filename of the OVA you deployed: v1.4.3-rc1-6053-e7623444
How was the OVA deployed? Flex client
Does the VIC appliance recieve configuration by DHCP? No
What stage of the Appliance Lifecycle is the VIC appliance in? Upgrade
VIC appliance support bundle:
[vic_appliance_logs_2018-09-11-13-48-40.tar.gz](https://github.com/vmware/vic-product/files/2371197/vic_appliance_logs_2018-09-11-13-48-40.tar.gz)
Old Version of VIC appliance:
`appliance=v1.3.1-3409-132fb13d
harbor=harbor-offline-installer-v1.3.0.tgz
engine=vic_v1.3.1.tar.gz
admiral=vmware/admiral:vic_v1.3.0 6f1d13decebe
vic-machine-server=gcr.io/eminent-nation-87317/vic-machine-server:latest 44e47c4b40ab
`
New Version of VIC appliance: 
`appliance=v1.4.3-rc1-6053-e7623444
harbor=harbor-offline-installer-v1.5.3-build.473.tgz
engine=vic_v1.4.3-rc1.tar.gz
admiral=vmware/admiral:vic_v1.4.3-rc1 b1416b4c9f6a
vic-machine-server=gcr.io/eminent-nation-87317/vic-machine-server:v1.4.3-rc1 5e949e13899a
`
Were the hard disks from the old appliance moved properly to the new appliance? YES
appliance upgrade script run log: 
[upgrade.log](https://github.com/vmware/vic-product/files/2371225/upgrade.log)
Was the new appliance initialized before running the appliance upgrade script? NO

Bug Report Detailed Information
What operation was being performed when the failure was noticed? Upgrade from 1.3.1 to 1.4.3 by manual method.

Steps Taken: Manually moved the disks from old VIC appliance to new VIC appliance as stated in the upgrade document.
Then executed:  `./upgrade.sh --manual-disks`
and enter yes when it asks to upgrade ui plugin

It hits an error when trying to upgrade flex and h5 plugin

Error: Upgrade failed.

lgayatri:
Log snippet:

```
set owner for /etc/vmware/psc/harbor/psc-config.keystore
[...]

PR #2053, opened 2018-09-11
Fix quoting of registration payload in upgrade.sh

zjs:
Because both the `payload` and `payload_info` variables were quoted with
single quotes, expressions were not being expanded.

Correct the quoting to properly print the JSON object; wrap constant
portions of the JSON with single quotes and expressions with double.

---

VIC Appliance Checklist:
- [ ] Added tests
   * I don't see a good way to do this, unfortunately.
- [x] Considered impact to upgrade
- [x] Tests passing
- [x] Updated documentation

Fixes: #2050 (see https://github.com/vmware/vic-product/issues/2050#issuecomment-420404212 for results of testing)

### DIFF
--- a/installer/build/scripts/upgrade/upgrade.sh
+++ b/installer/build/scripts/upgrade/upgrade.sh
@@ -97,8 +97,8 @@ function callPluginUpgradeEndpoint {
   local vc='{"target":"'"${VCENTER_TARGET}"'","user":"'"${VCENTER_USERNAME}"'","password":"'"${VCENTER_PASSWORD}"'","thumbprint":"'"${VCENTER_FINGERPRINT}"'"}'
   local vc_info='{"target":"'"${VCENTER_TARGET}"'","user":"'"${VCENTER_USERNAME}"'","thumbprint":"'"${VCENTER_FINGERPRINT}"'"}'
   local plugin='{"preset":"'"${preset}"'","force":true}'
-  local payload='{"vc":"${vc}","plugin":"${plugin}"}'
-  local payload_info='{"vc":"${vc_info}","plugin":"${plugin}"}'
+  local payload='{"vc":'"${vc}"',"plugin":'"${plugin}"'}'
+  local payload_info='{"vc":'"${vc_info}"',"plugin":'"${plugin}"'}'
   echo "register payload - ${payload_info}" >> $upgrade_log_file 2>&1
   /usr/bin/curl \
     -k \


installer/build/scripts/upgrade/upgrade.sh
```diff
@@ -97,8 +97,8 @@ function callPluginUpgradeEndpoint {
   local vc='{"target":"'"${VCENTER_TARGET}"'","user":"'"${VCENTER_USERNAME}"'","password":"'"${VCENTER_PASSWORD}"'","thumbprint":"'"${VCENTER_FINGERPRINT}"'"}'
   local vc_info='{"target":"'"${VCENTER_TARGET}"'","user":"'"${VCENTER_USERNAME}"'","thumbprint":"'"${VCENTER_FINGERPRINT}"'"}'
   local plugin='{"preset":"'"${preset}"'","force":true}'
-  local payload='{"vc":"${vc}","plugin":"${plugin}"}'
-  local payload_info='{"vc":"${vc_info}","plugin":"${plugin}"}'
+  local payload='{"vc":'"${vc}"',"plugin":'"${plugin}"'}'
+  local payload_info='{"vc":'"${vc_info}"',"plugin":'"${plugin}"'}'
```
renmaosheng: by looking at previous line, it seems we are quoting " ' ", which is different than current change.

zjs: That's intentional.

We're trying to build a JSON object like:

```json
{
  "vc": {
    "target": "...",
    "thumbprint":"...",
    "user":"...",
    "password":"..."
  },
  "plugin": {
    "preset":"flex",
    "force":true
  }
}
```

When we build `vc` and `plugin`, we want to put double quotes around the values in the generated output (as the properties are strings). When we build the top-level object, we omit the quotes (as the properties are objects).

In case it's not clear what the various quotes do:
```
local vc='{"target":"'"${VCENTER_TARGET}"'",...
         │ │      │ │││                 ││└── End a JSON property value string
         │ │      │ │││                 │└── Begin a bash literal
         │ │      │ │││                 └── End a bash literal containing an expression
         │ │      │ ││└── Begin a bash literal containing an expression
         │ │      │ │└── End a bash literal
         │ │      │ └── Begin a JSON property value string
         │ │      └── End a JSON property name
         │ └── Begin a JSON property name
         └── Begin a bash literal
```

```
local payload='{"vc":'"${vc}"',
              │ │  │ ││     │└── Begin a bash literal
              │ │  │ ││     └── End a bash literal containing an expression
              │ │  │ │└── Begin a bash literal containing an expression
              │ │  │ └── End a bash literal
              │ │  └── End a JSON property name
              │ └── Begin a JSON property name
              └── Begin a bash literal
```

renmaosheng: Thanks a lot!

Issue #2062, opened 2018-09-12
Document that you need to restart client after removing plugin from MOB


stuclem:
If you use the MOB procedure in https://vmware.github.io/vic-product/assets/files/html/1.4/vic_vsphere_admin/uninstall_vic.html to remove the plug-ins, you need to restart the vSphere Client. We should add this as a post-req.

Also, now that we have the auto-install, we should move the instructions for restarting the client out of http://localhost:4000/vic_vsphere_admin/ts_ui_not_appearing.html and into a separate topic.

PR #2089, opened 2018-09-14
Added step to restart client after removing plug-in

stuclem:
Fixes https://github.com/vmware/vic-product/issues/2062. @lgayatri since you just tested the MOB procedure, can you please review? Thanks!


stuclem:
Thanks @lgayatri

PR #2101, opened 2018-09-21
Removed note about FQDN

### LINKED ISSUE #323: Remove caveat about using registry IP in docker login
Per https://github.com/vmware/vic/issues/4721, remove the note from https://vmware.github.io/vic-product/assets/files/html/1.1/vic_app_dev/configure_docker_client.html and elsewhere that you cannot use the FQDN when running `docker login`.

vidya-v:
Hi @stuclem , I removed the note on FQDN from https://vmware.github.io/vic-product/assets/files/html/1.1/vic_app_dev/configure_docker_client.html 
Should I remove the line _The Create Virtual Container Host wizard only accepts an IP address for the client network. You cannot specify an FQDN._ from https://vmware.github.io/vic-product/assets/files/html/1.4/vic_vsphere_admin/client_network.html?

Fixes #323

### DIFF
--- a/docs/user_doc/vic_app_dev/configure_docker_client.md
+++ b/docs/user_doc/vic_app_dev/configure_docker_client.md
@@ -75,8 +75,7 @@ This example configures a Linux Docker client so that you can log into vSphere I
 
 This example configures a Linux Docker client so that you can log into vSphere Integrated Containers Registry by using its IP address.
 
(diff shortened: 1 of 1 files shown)


zjs:
> The Create Virtual Container Host wizard only accepts an IP address for the client network. You cannot specify an FQDN. from https://vmware.github.io/vic-product/assets/files/html/1.4/vic_vsphere_admin/client_network.html?

As far as I can tell, that line is still correct.

PR #2132, opened 2018-10-19
Replace toolbox with open-vm-tools

### LINKED ISSUE #916: Move appliance to open-vm-tools
**User Statement:**

As a developer I want to use open-vm-tools

**Details:**
https://reviewboard.eng.vmware.com/r/1171452/
Configure `tools.conf` with `exclude-nics=docker*`

**Acceptance Criteria:**
- [ ] Remove toolbox
- [ ] Use `open-vm-tools`

andrewtchin:
VIC Appliance Checklist:
- [x ] Up to date with `master` branch
- N/A Added tests
- [x ] Considered impact to upgrade
- [ ] Tests passing
- N/A Updated documentation
- [x ] Impact assessment checklist

Fixes #916

Toolbox was originally used because excessive Docker network and IPv6 addresses showed up in the vCenter UI and interfered with determining the actual ethernet interface IP. The update to open-vm-tools should have resolved that and according to that commit the default config should be good.
This needs to be built and deployed to verify that the desired IPv4 address shows up in the vCenter UI
There may also be a bugzilla for this, but I couldn't find it


installer/toolbox/main.go
```diff
@@ -1,89 +0,0 @@
...
-	service := toolbox.NewService(in, out)
-
-	if os.Getuid() == 0 {
-		service.Power.Halt.Handler = toolbox.Halt
-		service.Power.Reboot.Handler = toolbox.Reboot
-	}
-
-	// Disable all guest operations
-	service.Command.Authenticate = func(_ vix.CommandRequestHeader, _ []byte) error {
-		return errors.New("not authorized")
```
zjs: It seems like this behavior may still be desirable. Is there a way to preserve it with `open-vm-tools`?

hickeng: The ability to run guest operations can be controlled via vSphere RBAC.

hickeng:
The ask for this came from field facing staff who noted that they could not run Guest Customization due to toolbox not supporting that.

If we move to open-vm-tools we _must_ add testing to ensure that the OVA continues to function if guest customizations are performed via tools.

PR #2141, opened 2018-10-29
Adding source files for landing page at vmware.githib.io

stuclem:
These are the source files, taken from https://github.com/vmware/gh-page-template, that we use to generate the `vic-product` landing page at https://vmware.github.io/vic-product/. Storing these files in a `website` branch of the project repo, rather than in `gh-page-template` is the practice recommended by the open source Web team.


CODE_OF_CONDUCT.md
```diff
@@ -0,0 +1,19 @@
+Contributor Code of Conduct
+======================
+
+As contributors and maintainers of the Clarity project, we pledge to respect everyone who contributes by posting issues, updating documentation, submitting pull requests, providing feedback in comments, and any other activities.
```
zjs: It looks like this may be part of the template, but probably not something we want to copy into this repo as-is. (We should have a code of conduct, but I think that's a separate, in-progress task.)

CONTRIBUTING.md
```diff
@@ -0,0 +1,113 @@
+
+
+# Contributing to clarity-seed
```
zjs: I think we can remove this for now.

Issue #2111, opened 2018-09-27
Create a testbed for upgrade and modify test cases.


renmaosheng:
**User Statement:**
current vic-product nightly testing for upgrade (group 7 and group 8), the ESX resource is not enough to run two ova on the ESX host, which causes the testing not stable,  https://vic-jenkins.eng.vmware.com/view/Scenario/job/vic-product-release-scenario-test/ 

**Acceptance Criteria:**
the vic-product nightly testing for upgrade (group 7 and group 8) pass consistently

PR #2169, opened 2018-11-08
Create Keyword to use rubyfile to create 2 esxi with shared iscsi lun as test env

luwang-vmware:
Create Keyword to use rubyfile to create 2 esxi with shared iscsi lun as test env

Part of fix issue #2111

Create Keyword Setup Simple VC And Test Environment with Shared iSCSI Storage
Create Keyword Create Simple VC Cluster With Spec File
Update test env in Group7-Upgrade cases to use this new test env


luwang-vmware:
Attached pass log for Group7 with the code change. https://vic-jenkins.eng.vmware.com/view/LuWang-Test/job/luwang-vic-product/15/

Issue #2181, opened 2018-11-23
Clarify whether restarting the VCH manually is a problem


stuclem:
Per internal bug:

A question came up on a section in the documentation. 
" Manually restarting container VMs can result in incorrect end-times for container operations. Do not manually restart the VCH or container VMs. Always use Docker commands or the vSphere Integrated Containers Management Portal to perform operations on container VMs."
https://vmware.github.io/vic-product/assets/files/html/1.3/vic_vsphere_admin/interop.html 

It is pretty regular to restart the VCH from the vSphere Web client and I do not recall a way to do so from the container management portal. Should the VCH be included in this warning “Do not manually restart the VCH or container VMs”? If this statement is correct to include the VCH, then it might be helpful to explain a little about what conditions the VCH might be restarted manually for why the bullet point “If you restart a VCH endpoint VM, it comes back up in the same state that it was in when it shut down” was made for. It is easy to read these as if they contradict each other without knowing that there are exceptions to the general rule for restarting the VCH.

PR #2182, opened 2018-11-23
Removed instruction not to manually restart VCH

stuclem:
Fixes https://github.com/vmware/vic-product/issues/2181.

I think that the instruction not to restart VCHs from the vSphere Client was either a hangover from an old problem that has since been eliminated, or else a flat-out error. @zjs can you please confirm?


hickeng:
@stuclem I *think* this should be fixed for containerVMs as well with #6943 (will be in release 1.5). This adds support for the `guest shutdown` and `guest restart` commands that perform a polite power operation.

The issue will still be present if a customer uses the hard `power off` or `reset` operations as there is no opportunity for the tether to update the timestamps.

PR #2195, opened 2018-11-29
Verify vic password during (re)initialization

wjun:
Fixes #[93](https://github.com/vmware/vic-tasks/issues/93)

### DIFF
--- a/installer/build/scripts/verify.py
+++ b/installer/build/scripts/verify.py
@@ -1,4 +1,4 @@
-#!/usr/bin/python
+#!/usr/bin/env python3
 # Copyright 2018 VMware, Inc. All Rights Reserved.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
@@ -15,22 +15,19 @@
 
 
 import crypt
-import subprocess
+import spwd
 import sys
+from hmac import compare_digest
 
 
 def main():
     if len(sys.argv) != 2:
         exit(1)
-    salt = subprocess.check_output(
-        "sudo getent shadow root | cut -d$ -f3",
-        shell=True).rstrip()
-    origin_passwd = subprocess.check_output(
-        "sudo getent shadow root | cut -d: -f2",
-        shell=True).rstrip()
-    passwd = crypt.crypt(sys.argv[1], "$6$" + salt)
-    if origin_passwd != passwd:
+    cryptedpasswd = spwd.getspnam("root")[1]
+    if not compare_digest(crypt.crypt(sys.argv[1], cryptedpasswd),
+                          cryptedpasswd):
         exit(1)
+
 
 if __name__ == "__main__":
     main()
--- a/installer/build/scripts/verify.py
+++ b/installer/build/scripts/verify.py
@@ -1,4 +1,4 @@
-#!/usr/bin/bash
+#!/usr/bin/python
 # Copyright 2018 VMware, Inc. All Rights Reserved.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
@@ -13,7 +13,25 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-salt=$(sudo getent shadow root | cut -d$ -f3)
-origin_password=$(sudo getent shadow root | cut -d: -f2)
-passwd=$(python -c 'import crypt; print crypt.crypt("'"${1}"'", "$6$'"${salt}"'")')
-[ "${origin_password}" == "${passwd}" ] && exit 0 || exit 1
+
+import crypt
+import subprocess
+import sys
+
+
+def main():
+    if len(sys.argv) != 2:
+        exit(1)
+    salt = subprocess.check_output(
+        "sudo getent shadow root | cut -d$ -f3",
+        shell=True).rstrip()
+    origin_passwd = subprocess.check_output(
+        "sudo getent shadow root | cut -d: -f2",
+        shell=True).rstrip()
+    passwd = crypt.crypt(sys.argv[1], "$6$" + salt)
+    if origin_passwd != passwd:
+        exit(1)
+
+if __name__ == "__main__":
+    main()
+
--- a/installer/build/scripts/verify.sh
+++ b/installer/build/scripts/verify.sh
@@ -1,0 +1,25 @@
+#!/usr/bin/bash
+# Copyright 2018 VMware, Inc. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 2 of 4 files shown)


installer/fileserver/routes/index.go
```diff
@@ -75,7 +77,16 @@ func (i *IndexHTMLRenderer) IndexHandler(resp http.ResponseWriter, req *http.Req
 
 // indexFormHandler registers the appliance using post form values
 func indexFormHandler(op trace.Operation, req *http.Request, html *IndexHTMLOptions) error {
-	// verify login
+	// verify vic appliance root password
+	vicPasswd := req.FormValue("appliancePwd")
+	cmd := exec.Command("/etc/vmware/verify.sh", vicPasswd) // #nosec
```
zjs: We should have a security expert comment on this approach.

wjun:
@zjs The UI changes have been committed first at https://github.com/vmware/vic-product/commit/acb7094f6f5bbd25dc7e0033cfe88da1dd044b29 . For the security check, I only use password input to python crypt api in verify.sh.

vmlaguna:
Hi all,

vicPasswd is passed without filtering or validation directly into a script that is execute as root. This script later uses this value without correctly handling it, allowing for command execution.

I will like to have a defense in deep in this approach:

- index.go: If we currently have some limitations and validations for passwords (min 8 characters, must include capitals, do not include special characters, etc.) we should validate that before sending the string to the script.
- verify.sh: Use tools like Shellcheck to ensure script is up to good scripting standards. Ensure input parameter `first_param` is treated as literal and can not be used as a way to inject parameters.

There is also the problem of a timing attack against the password comparison but as we are comparing a sha512 hash that later the attacker will have to crack I see this as a very low risk issue: https://crypto.stackexchange.com/questions/25607/practical-uses-for-timing-attacks-on-hash-comparisons-e-g-md5

Please, update both files with the recommendations and ping me in case you need some help with this.

wjun:
Thanks @vmlaguna for the comments. I rewrote verify.sh with shellcheck tool.

installer/build/scripts/verify.sh
```diff
@@ -0,0 +1,19 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+salt=$(sudo getent shadow root | cut -d$ -f3)
+origin_password=$(sudo getent shadow root | cut -d: -f2)
+passwd=$(python -c 'import crypt; print crypt.crypt("'"${1}"'", "$6$'"${salt}"'")')
```
zjs: This method of passing the password to python protects against bash evaluation attacks, but still allows for injection.

For example, if the value of `$1` is `", ""); import os; print os.system("sudo touch /this-is-bad"); #` then the file `this-is-bad` will be created in the root directory. This command could be replaced with an arbitrary command, allowing an unauthenticated user to perform an operation on the appliance as `root`.

wjun: This is a good catch! I should not evaluate the input. Changed the shell script to python script to avoid input evaluation.

installer/fileserver/routes/index.go
```diff
@@ -75,7 +77,16 @@ func (i *IndexHTMLRenderer) IndexHandler(resp http.ResponseWriter, req *http.Req
 
 // indexFormHandler registers the appliance using post form values
 func indexFormHandler(op trace.Operation, req *http.Request, html *IndexHTMLOptions) error {
-	// verify login
+	// verify vic appliance root password
+	vicPasswd := req.FormValue("appliancePwd")
+	cmd := exec.Command("/etc/vmware/verify.py", vicPasswd) // #nosec
```
vmlaguna: Why are we adding the #nosec here?

installer/build/scripts/verify.py
```diff
@@ -0,0 +1,37 @@
+#!/usr/bin/python
```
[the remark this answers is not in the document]
vmlaguna: Also, recommended to use a tool like bandit to review the script: https://bandit.readthedocs.io/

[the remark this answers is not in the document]
wjun: Both are very good suggestions. Fixed to use env and use bandit to scan.

installer/build/scripts/verify.py
```diff
@@ -0,0 +1,37 @@
...
+    if len(sys.argv) != 2:
+        exit(1)
+    salt = subprocess.check_output(
+        "sudo getent shadow root | cut -d$ -f3",
+        shell=True).rstrip()
+    origin_passwd = subprocess.check_output(
+        "sudo getent shadow root | cut -d: -f2",
+        shell=True).rstrip()
+    passwd = crypt.crypt(sys.argv[1], "$6$" + salt)
+    if origin_passwd != passwd:
```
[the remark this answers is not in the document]
wjun: Done

installer/fileserver/routes/index.go
wjun: We enabled Golang gas for code security scan, and Subprocess launching with variable. (Confidence: HIGH, Severity: HIGH) for this statement, which is true as we code reviewed here. I need to add this tag to pass gas check during the ci build, if we make sure verify.py is secure.

installer/fileserver/routes/index.go
```diff
@@ -75,7 +77,16 @@ func (i *IndexHTMLRenderer) IndexHandler(resp http.ResponseWriter, req *http.Req
 
 // indexFormHandler registers the appliance using post form values
 func indexFormHandler(op trace.Operation, req *http.Request, html *IndexHTMLOptions) error {
-	// verify login
+	// verify vic appliance root password
+	vicPasswd := req.FormValue("appliancePwd")
```
vmlaguna: One last check I will like to see implemented is some sort of validation on the password string. We shouldn't limit the charset as we want users to use complex passwords but do we have a limit on how many characters the password needs to be? Let's say we only accept passwords up to 256 characters, we should enforce this here.

I am trying to think on as many checks we can implement without breaking the functionality to ensure we do not leave anything open.

installer/fileserver/routes/index.go
```diff
@@ -75,7 +79,22 @@ func (i *IndexHTMLRenderer) IndexHandler(resp http.ResponseWriter, req *http.Req
...
+	if len(vicPasswd) > 30 {
```
zjs: We should ensure this limit is documented.

cc @stuclem

vmlaguna: This should not be a hardcoded value but something that comes from the configuration of the system.

PR #2221, opened 2018-12-07
Modify ways to identify host from css to xpath by host IP

### LINKED ISSUE #2220: Fix random failure of case 5-01-Distributed-Switch-No-Cluster.robot
5-01-Distributed-Switch-No-Cluster.robot is randomly failure in scenario testing. The root cause is that our case is using Nimbus to create 3 ESXi which all have datastore1 and when they are added to VC, the 2nd and 3rd hosts' datastore1 is named as datastor1(*). VIC UI sorted the 3 ESXi host with their ip address and our case are hardcode to use the 1st host as VCH compute resource, but sometimes the 1st host is not the one which is the 1st added to VC, so case get failure as no 'datastore1' found during picking up image datastore step.

danfengliu:
Due to GUI design of listing VCHs was changed to sort by IP address, so script '5-01-Distributed-Switch-No-Cluster' failed to locating the wanted VCH, it wants a VCH with property 'datastore' named 'datastore1', it always at the head of VCH nodes.
After design changed, the position of this VCH may not be at the head of the list, so add a 'if condition' in keyword 'Create VCH using UI And Set Docker Parameters' to locate VCH index by '1' or IP address.

Fixes #2220

### DIFF
--- a/tests/resources/page-objects/Vsphere-VCH-Plugin-Util.robot
+++ b/tests/resources/page-objects/Vsphere-VCH-Plugin-Util.robot
@@ -174,8 +174,12 @@ Create VCH using UI And Set Docker Parameters
     # compute capacity
     Log To Console  Selecting compute resource...
     # if cluster is present
-    Wait Until Element Is Visible And Enabled  css=.clr-treenode-children clr-tree-node:nth-of-type(${tree-node}) .cc-resource
-    Click Button  css=.clr-treenode-children clr-tree-node:nth-of-type(${tree-node}) .cc-resource
+    # There're 2 types of tree-node, if it's a index number, locator will be defined by CSS, if it's a IP address, locator will be defined by xpath 
+    ${tree-node-len}=  Get Length  '${tree-node}'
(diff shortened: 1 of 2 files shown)


tests/resources/page-objects/Vsphere-VCH-Plugin-Util.robot
```diff
@@ -172,10 +172,12 @@ Create VCH using UI And Set Docker Parameters
     Input VCH Name  ${name}
     Click Next Button
     # compute capacity
-    Log To Console  Selecting compute resource...
     # if cluster is present
-    Wait Until Element Is Visible And Enabled  css=.clr-treenode-children clr-tree-node:nth-of-type(${tree-node}) .cc-resource
-    Click Button  css=.clr-treenode-children clr-tree-node:nth-of-type(${tree-node}) .cc-resource
+    ${tree-node-len}  Get Length  '${tree-node}'
+    ${host_locator}  set variable if  ${tree-node-len}<5  css=.clr-treenode-children clr-tree-node:nth-of-type(${tree-node}) .cc-resource  xpath://clr-tree-node//button[contains(., '${tree-node}')]
```
luwang-vmware: We had better to add comments here.
when tree-node is host ip address, scripts uses xpath to select it.

danfengliu: Yes, I've added one.

Issue #2225, opened 2018-12-10
Document maximum length of VIC appliance password


stuclem:
Per comment from @zjs in https://github.com/vmware/vic-product/pull/2195#ref-pullrequest-388221853, we need to document the max password length.

PR #2226, opened 2018-12-10
Documented that appliance password is limited to 30 chars

stuclem:
Fixes https://github.com/vmware/vic-product/issues/2225.

@wjun is this correct? Thanks!


stuclem:
Thanks @wjun!

PR #2206, opened 2018-12-04
Updating version/year for 1.5, removing plug-in docs

stuclem:
This PR does the following:

- Updates the version to 1.5
- Updates the copyright date to 2019
- Removes all reference to the script-based installation and upgrade of the UI plug-in (needed to be retained in the 1.4.x docs for releases prior to 1.4.3, but now redundant in 1.5)
- Removes notes about features that were added in 1.4.3

PR #2236, opened 2018-12-14
Fix incorrect plugin api usage

wjun:
vic appliance password is part of appliance object.


wjun:
@stuclem  Stuart, I do not know if we document apis to install/remove/upgrade UI plugins or not. If so, we also need to update that part to add vicpassword parameter as this patch shows. Thanks.

stuclem:
@wjun, starting with 1.5.0 we only document the automatic install/upgrade of the UI plug-ins. We have removed the manual method in this release, as this was only applicable until 1.4.3. 

See https://github.com/vmware/vic-product/pull/2206 for details. Thanks!

Issue #2159, opened 2018-11-06
Document --storage-quota options


stuclem:
Per https://github.com/vmware/vic-planning/issues/43. Applies to both `vic-machine create` and `vic-machine configure`.

stuclem:
Per https://github.com/vmware/vic/issues/8313, the storage quota is also shown in `docker info`.

stuclem:
The `--storage-quota` option is implemented in https://github.com/vmware/vic/pull/8327.

stuclem:
Doesn't appear in UI.

PR #2271, opened 2018-12-27
Documented --storage-quota option

stuclem:
Fixes https://github.com/vmware/vic-product/issues/2159.

@wjun is this OK? Thanks!


docs/user_doc/vic_vsphere_admin/vicmachine_options_ref.md
```diff
@@ -66,6 +66,7 @@ This topic includes lists of all of the options of the different `vic-machine` c
 |`--public-network-gateway`|[Configure the Public Network](public_network.md#gateway)|
 |`--public-network-ip`|[Configure the Public Network](public_network.md#static-ip)|
 |`--registry-ca`|[Configure Registry Access](vch_registry.md#registry-ca)|
+|`--storage-quota`|[Specify the Image Datastore](image_store.md#quota)|
```
wjun: Better to change the description to "Specify the Image Datastore Limit"?

stuclem: Hi @wjun this is actually the title of the topic/section in which the option is documented, rather than a description. So, it should stay as it is.

Are the changes OK otherwise? Thanks!

PR #2284, opened 2019-01-08
Add logging to appliance init

### LINKED ISSUE #777: Add logging to appliance init
**User Statement:**

As a person I want better logs in the Getting Started Page's appliance initialization.

**Details:**
Also make sure that the registration API is covered
https://github.com/vmware/vic-product/blob/master/installer/fileserver/server.go#L244

**Acceptance Criteria:**
- [ ] Add logging

DanielXiao:
Write logs of ova-webserver to /var/log/fileserver.

VIC Appliance Checklist:
- [ ] Added tests
- [ ] Tests passing
- [ ] Updated documentation

Fixes #777


installer/fileserver/main.go
```diff
@@ -83,11 +85,23 @@ func parseServerConfig(op trace.Operation, conf *serverConfig) {
 		trace.Logger.Level = log.DebugLevel
 	}
 
+	var err error
+
+	if _, err := os.Stat(conf.logPath); os.IsNotExist(err) {
+		os.MkdirAll(conf.logPath, 0700)
+	}
```
YanzhaoLi: I thinks we should also check whether `conf.logPath` is a directory.  Because users may provide a filename path to `conf.logPath`.

installer/fileserver/main.go
```diff
@@ -83,11 +85,23 @@ func parseServerConfig(op trace.Operation, conf *serverConfig) {
...
+	logPath := filepath.Join(conf.logPath, "fileserver.log")
+	logFile, err := os.OpenFile(logPath, os.O_RDWR|os.O_APPEND|os.O_CREATE, 0600)
```
YanzhaoLi: Do we need to handle closing the logFile, or the trace.Logger.out already help us handle it?

installer/fileserver/main.go
DanielXiao: MkdirALL will return Error if it is a file
```
		if dir.IsDir() {
			return nil
		}
		return &PathError{"mkdir", path, syscall.ENOTDIR}
```

installer/fileserver/main.go
DanielXiao: We can't close the file descriptor because the running process will write logs to it until it exits.

DanielXiao:
@zjs Do you think it is necessary to write logs to a file?

zjs:
I think #777 is about improving the _content_ of the logs, not changing how they're written (although that may be useful too, as a vSphere Administrator attempting to debug may be more used to finding files in `/var/log` than checking a journal — but in that case I think we'd want to write them both ways).

For example, I believe we don't currently get good logging around the plugin registration process or failures that can occur there.

DanielXiao:
@wjun From logging design principle https://github.com/vmware/vic-product/blob/master/installer/docs/DESIGN.md#logging, the init log should be written to a file.

Issue #2247, opened 2018-12-17
Document that new appliance pw is required during upgrade


stuclem:
Per https://github.com/vmware/vic-product/issues/2237#issuecomment-447786516

PR #2302, opened 2019-01-10
Added step to provide new appliance password to upgrade doc

stuclem:
Fixes https://github.com/vmware/vic-product/issues/2247. @wjun can you please review? Thanks!


stuclem:
Thanks @wjun !

PR #2311, opened 2019-01-16
Updated landing page and regen HTML again

stuclem:
Generated HTML. Source files are in master and website branches.


stuclem:
Thanks @vidya-v !

Issue #2368, opened 2019-03-05
Document that you can set an absolute cVM limit per VCH


stuclem:
Per https://github.com/vmware/vic/issues/6529

PR #2370, opened 2019-03-05
Documented --containers option

stuclem:
Fixes https://github.com/vmware/vic-product/issues/2368. 

@wjun and @yuyangbj can you please review? Thanks!

### DIFF
--- a/docs/user_doc/vic_vsphere_admin/configure_vch.md
+++ b/docs/user_doc/vic_vsphere_admin/configure_vch.md
@@ -188,7 +188,7 @@ If you set a new limit on a VCH that is lower than the number of container VMs t
 
 **NOTE**: This option is available in vSphere Integrated Containers 1.5.2 and later.
 
-The `vic-machine configure --containers` option functions in the same way as the equivalent `vic-machine create` option. For information about the `vic-machine create --containers` option, see [Container VM Limit](image_store.md#container-limit) in Specify the Image Datastore.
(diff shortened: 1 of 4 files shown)


stuclem:
@wjun and @yuyangbj, I added the `vic-machine create --containers` option to the topic on configuring the image store, so that it is alongside the `--storage-quota` option. However, I'm not sure that this is the correct place. The docs describing the `vic-machine create` options match the layout of the Create VCH wizard. So, do you know where the container VM limit will appear when we implement it in the UI? I'm thinking that it could be in any of the following places:

- the General page, alongside the Container VM name template option
- the Storage Capacity page, alongside storage quota
- the Configure Networks page, in the section on container networks. 

What do you think? If we know where in the UI it will go, it will be best to match that placement in the docs now, rather than having to move the doc after we update the UI. Thanks!

wjun:
> @wjun and @yuyangbj, I added the `vic-machine create --containers` option to the topic on configuring the image store, so that it is alongside the `--storage-quota` option. However, I'm not sure that this is the correct place. The docs describing the `vic-machine create` options match the layout of the Create VCH wizard. So, do you know where the container VM limit will appear when we implement it in the UI? I'm thinking that it could be in any of the following places:
> 
> * the General page, alongside the Container VM name template option
> * the Storage Capacity page, alongside storage quota
> * the Configure Networks page, in the section on container networks.
> 
> What do you think? If we know where in the UI it will go, it will be best to match that placement in the docs now, rather than having to move the doc after we update the UI. Thanks!

@stuclem I prefer to the general page section, as it is a resource limit to the VCH, but not related to storage or network only. The contents are accurate. Thanks.

stuclem:
Thanks @wjun. I have moved the container VM limit option to the General VCH Settings page.

Issue #2365, opened 2019-03-05
Document that changing the client network gateway is optional in UI


stuclem:
Per https://github.com/vmware/vic-ui/pull/655.

PR #2371, opened 2019-03-05
Documented that client gateway is not mandatory in wizard

stuclem:
Fixes https://github.com/vmware/vic-product/issues/2365.

@pureshine and @zhoumeina can you please review? Thanks!


stuclem:
Thanks @pureshine !

PR #2382, opened 2019-03-12
Regen HTML with attempt to fix Google Analytics

stuclem:
Generated HTML files. Sources are in master.


stuclem:
Thanks @vidya-v !

PR #2565, opened 2020-05-26
Disable tls1.1 when users access to 9443 port

ading007:


installer/lib/tls.go
```diff
@@ -22,7 +22,7 @@ func GetTLSServerConfig(cert tls.Certificate) *tls.Config {
 	// forcing tls 1.1, cipher from https://github.com/denji/golang-tls#perfect-ssl-labs-score-with-go
```
renmaosheng, line 22: could you please also update the comment to tls 1.2? thanks.

PR #2619, opened 2022-07-08
Remove /installer/fileserver/test directory

ading007:
We need to remove the directory to fix security issue for npm package.
Due to we will not add feature for vic product UI, so the test cases isn't
necessary.


xuelichao (approved):
LGTM

